PR #50, opened 2022-08-14
Noetic debug propulsion

ben-jamming:
Fixed issues with pool-test package including:

1. Synchronization errors between Arduino and Jetson Nano by standardizing baud rate to be 57600.
2. Modified pin configuration in thrusters.ino to lineup with ThusterCommands.msg
3. Updated pool-test/src/thrusters.py to display the correct pins when testing.


catkin_ws/src/propulsion/embedded/thrusters/thrusters.ino
```diff
@@ -47,14 +53,14 @@ uint16_t* convertToMicroseconds(const float intensities[8]){
...
 
-	thrusters[SRG_P].writeMicroseconds(command[SRG_P_PIN]);
-	thrusters[SRG_S].writeMicroseconds(command[SRG_S_PIN]);
-	thrusters[SWY_BW].writeMicroseconds(command[SWY_BW_PIN]);
-	thrusters[SWY_ST].writeMicroseconds(command[SWY_ST_PIN]);
-	thrusters[HVE_BW_P].writeMicroseconds(command[HVE_BW_P_PIN]);
-	thrusters[HVE_BW_S].writeMicroseconds(command[HVE_BW_S_PIN]);
-	thrusters[HVE_ST_P].writeMicroseconds(command[HVE_ST_P_PIN]);
-	thrusters[HVE_ST_S].writeMicroseconds(command[HVE_ST_S_PIN]);
+	thrusters[SRG_P].writeMicroseconds(command[SRG_P_PIN-2]);
```
VasilyF, line 56: use SRG_P instead

catkin_ws/src/dry-test/src/thrusters.py
```diff
@@ -3,10 +3,11 @@
...
 from auv_msgs.msg import ThrusterCommand
 
-pins = {1:4, 2:5, 3:6, 4:7, 5:8, 6:9, 7:10, 8:11}
-reset = [0, 0, 0, 0, 0, 0, 0, 0]
+pins = {1:2, 2:3, 3:4, 4:5, 5:6, 6:7, 7:8, 8:9}
+reset = [0,0,0,0,0,0,0,0]
 reset_cmd = ThrusterCommand(reset)
-pub = rospy.Publisher('propulsion/thruster_cmd', ThrusterCommand, queue_size=50)
+pub = rospy.Publisher('propulsion/thruster_cmd', ThrusterCommand, queue_size=10)
+rospy.sleep(7)
```
VasilyF, line 11: why sleep here?

catkin_ws/src/propulsion/embedded/thrusters/thrusters.ino
```diff
@@ -77,7 +83,7 @@ void initThrusters(){
 	thrusters[HVE_ST_P].attach(HVE_ST_P_PIN);
 
 	//set initial thruster effort (OFF)
-	thrustersOff();
+	//thrustersOff();
```
VasilyF, line 86: what was the issue with setting the thrusters to initially be off?

catkin_ws/src/dry-test/src/thrusters.py
rsaif22: To initialize the ESCs. When we create a new publisher, it initializes the topic with all zero values which for some reason re-initializes the ESCs. I put it in to avoid that for now. It could be that the current is too low, which I'll check tomorrow.

catkin_ws/src/propulsion/embedded/thrusters/thrusters.ino
rsaif22: That's the initialization step. Setting it to zero here and then creating a publisher to the topic makes it re-initialize and lose sync.

PR #64, opened 2022-10-22
Tutorial 1 exercise jana

jana-strizak:


VasilyF:
Nice work! Couple of points:
- Careful what branch you are on when you create the `tutorial-1-exercise-Jana` branch. Here, the commit from `tutorial-1-Jana` was also included in the pull request (not sure if you meant to include it). If you know what branch you will be merging (pull requesting) into you should originally branch off of there
- Don't be afraid to delete stuff - we can always recover things from looking at the git history if need be
- The example changes in `thrust_mapper.py` are not quite right - you would encounter an error when you run the file: `NameError: name 'wrench_to_thrust' is not defined` because in the code `wrench_to_thrust` exists only within the scope of the `effort_cb` function

PR #70, opened 2022-10-29
use update_effort as subscriber callback

ananemyro:


VasilyF:
I checked and the indentation used is: 1 level of indentation = 4 spaces

PR #73, opened 2022-10-31
Tutorial 2 antoine dangeard

antoinedang:


VasilyF:
everything looks good, it's definitely better to calculate the required PWM based on a quadratic relationship - https://bluerobotics.com/store/thrusters/t100-t200-thrusters/t200-thruster-r2-rp/

Right now, this would need to happen inside of thruster.ino because that relevant conversion is happening there (line 48). In the future we could just do all these calculations in Python and just pass the desired PWM

PR #75, opened 2022-11-10
Tutorial 2 Exercise: Implementing PID controller (only proportional term)

chanimores:
Hello, here's my PR for Tutorial 2. The first commit consists of the content of the tutorial itself, while the second commit is the code changes needed for the exercise. `pid.py` subscribes to two new topics, `/setpoint_x` and `/state_x`, wherein it will update the setpoint if necessary, and upon a new state being posted, it will publish the required effort to the topic `/surge` to get to that specified point.


VasilyF:
LGTM

PR #76, opened 2022-11-13
Changed state estimation to publish state values to seperate topics, …

MrMondrian:
…take data from the depth sensor and publish the negative value, and pid.launch in controls listends to state_z and publishes to /heave. Note you need to set the setpoint for pid to work


catkin_ws/src/state_estimation/launch/state_estimation.launch
```diff
@@ -2,6 +2,7 @@
 	<include file="$(find depth_sensor)/launch/depth_sensor.launch"/>
 	<!-- 
     for the current pool tests we are publishing just /depth directly from the sensor/rosserial node
```
VasilyF, line 4: delete the entire comment since it no longer applies - otherwise fine to merge

VasilyF: Merging and deleting the comment in a new commit

PR #77, opened 2022-11-16
Logarithmic pwm

antoinedang:
I haven't changed any arduino code, since I believe right now that it linearly maps an input intensity (-1 -> 1) to the motor thrust value (from min to max). My code follows the same template as in tutorial 2 where the new node subscribes to the topic where the thrust mapper publishes intensities (in Newtons), and publishes the corresponding intensities to the arduino intensities topic. Also added a stiffness hyperparameter which can be increased to make the logarithm closer to a unit step function


VasilyF:
The changes should be merged into noetic-stability.

It seems you made these commits off of an older version of the noetic branch and now there's a few conflicts with more recent changes.

I think the easiest remedy would be to create a new branch off of noetic-stability. Make a new commit and pull request the new branch:
```
git checkout noetic-stability
git pull
git checkout -b noetic-stability-logarithmic-pwm
# redo changes
```

PR #79, opened 2022-11-20
Logarithm pwm stability

antoinedang:
Same code and functionality as before but started from correct branch

### DIFF
--- a/catkin_ws/src/propulsion/src/force_to_intensity.py
+++ b/catkin_ws/src/propulsion/src/force_to_intensity.py
@@ -6,20 +6,17 @@ from auv_msgs.msg import ThrusterForces, ThrusterIntensities
 
 
 # forces produced by T200 thruster at 16V (N)
-MAX_FWD_FORCE = 5.25*9.81 
-MAX_BKWD_FORCE = 4.1*9.81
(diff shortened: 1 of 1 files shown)


VasilyF:
- I think that to go from force -> intensity you need to take the square root, not log2. 
consider a desired force of 9.81 N (1.0 kg f). the graph in the [docs](https://bluerobotics.com/store/thrusters/t100-t200-thrusters/t200-thruster-r2-rp/)  suggest that this occurs at ~1640 PWM (35% intensity) however this program outputs 69% intensity. 
- Our current battery is closer to 14 V -> (MAX_FWD: 4.52 kg f, MAX_BACK: 3.52 kg f)

antoinedang:
Oops my bad, should be fixed now

VasilyF:
The code is correct for modelling the relationship as quadratic as far as I can tell, however there still appears to be a discrepancy with that graph in the technical docs:

for forward 9.81 N (1 kg f), 14V:
- our value : 47%
- documented value: 38%

I'm merging the changes now. It would be good to later determine a more accurate model based on the [empirical data](https://cad.bluerobotics.com/T200-Public-Performance-Data-10-20V-September-2019.xlsx)

PR #80, opened 2022-11-23
Add Heave Calibration node and Calibration Package

chanimores:
For the Heave Eccentricity Calibration task:

- Added the `Calibration` package
- Added a `heave-calibration` node that sets the `theta_y` setpoint to `0.0` to level the AUV compared to the y-axis- PID will pitch the AUV until it meets the setpoint
- Added a `theta_y` PID to do this rotation

`calibration.launch`  allows us to calibrate the thrusters so the AUV remains upright while underwater without rotation due to any eccentricities due to mass/volume


VasilyF:
looks good thanks!

PR #81, opened 2022-11-27
Added other orientation calibration nodes and corresponding orientation PIDs

chanimores:
- Renamed `heave_calibration` to `nominal_calibration`
- Added two more calibration nodes (one where pitch is 90 degrees and the other where roll is 90 degrees) and their corresponding launch files

Note: this assumes that `theta_x` corresponds to `roll`, `theta_y` corresponds to `pitch`, and `theta_z` corresponds to `yaw.` I wasn't 100% sure on these orientations but these were the ones that made the most sense to me.  

Also important to note the PID setpoint messages are also sent at the same time.


VasilyF:
awesome thanks!

PR #82, opened 2023-01-07
Force to microseconds

antoinedang:
I curve-fitted the thrust t200 data (one curve for positive forces, one for negative) to 7-degree polynomials (seemed like the best tradeoff for std. error to degree, curve fit description in .png files in the propulsion folder), and made it so that Wrenches from the /effort topic are converted into microseconds using these polynomials directly in the thrust_mapper node. They're then sent to the propulsion/thruster_microseconds topic, and I modified the arduino code to read from that topic and use those values directly. Builds normally (requires a catkin clean first), and I tested both polynomials and the whole node.


VasilyF:
Awesome, thanks so much! I just removed the generated images and the extra CMakeLists.txt file. 

I haven't tested the arduino code, but the changes seem straightforward. It would be good to test that though and simultaneously update the thrusters test in the dry-test package (which publishes messages to the arduino directly).

I did notice that for small force values (ie. -2e-15 that could be generated due to machine precision errors when converting effort into thruster force) it will spit out a PWM of 1470 - this works because there's a band around 1500 where PWM values map to 0 N, it was just a bit confusing the first time I saw it, though it shouldn't matter practically.

To replicate: 
```
rostopic pub -1 /effort geometry_msgs/Wrench "force:
  x: 100.0
  y: 0.0
  z: 0.0
torque:
  x: 0.0
  y: 0.0
  z: 0.0" 

```

antoinedang:
Ok no worries, I'll check out the dry test code in time for the dry test next saturday

PR #85, opened 2023-01-13
Imu

MrMondrian:
Imu package is working and publishing data. It reads off ttyACM0 and publishes to a topic called IMU_DATA in the form of an imu message which is 3 floats, roll pitch and yaw.


VasilyF:
Awesome thanks!

PR #87, opened 2023-01-21
State Controller task

VasilyF:


VasilyF:
merging it for now, we can fix it up later

PR #90, opened 2023-01-31
added angle modulo code, untested

antoinedang:
Should be simple to test/adjust values if necessary, I think the code in itself is correct though from just static analysis (I'm, assuming imu doesn't jump around in value it outputs unless it reaches a modulo limit)


VasilyF:
We can test it on Thursday to get the actual modulo behaviour and merge it in thereafter

PR #91, opened 2023-02-06
refactor thrust_mapper

ben-jamming:


VasilyF:
I think you can put MAX_FWD_FORCE in thrust_mapper_utils since it's not really used inside thrust_mapper now.

Also, we should look into the TODO regarding syncing with Arduino - I'm not sure whether that's an issue or not

PR #96, opened 2023-02-23
superimposer acts in robot's reference frame

averystraumann:
Removed reference frame translations from superimposer.py; assumed that inputs will already be in the robot's reference frame, so there's no need to transform them


VasilyF:
Nice work!

PR #101, opened 2023-03-03
fixes to thrustmapper

MrMondrian:


VasilyF:
seems fine - though later we can edit the utils file to contain the MAX constants

PR #108, opened 2023-03-24
update joystick mappings

averystraumann:
Updated joystick mappings and created a node to increase/decrease heave.
New mappings:

deadman button: [5] (right bumper); must be held for joystick inputs to be registered
sway: axis 0 (left joystick, horizontal axis)
surge: axis 1 (left joystick, vertical axis)
roll: axis 2 (left trigger)
yaw: axis 3 (right joystick, horizontal axis)
pitch: axis 4 (right joystick, vertical axis)
heave, +z: button 1 (red circle)
heave, -z: button 0 (blue X)

### DIFF
--- a/catkin_ws/src/controls/config/joy_teleop.yaml
+++ b/catkin_ws/src/controls/config/joy_teleop.yaml
@@ -3,7 +3,7 @@ teleop:
     type: topic
     message_type: std_msgs/Float64
     topic_name: surge
-    deadman_buttons: [8]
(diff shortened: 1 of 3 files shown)


VasilyF:
Everything seems like it would work fine. Some nitpicks about joy_heave:
- you can use the same publisher to publish positive/negative messages to heave
- you can publish directly inside the if statement rather than setting the state of a boolean and then deciding whether to publish based on the boolean

PR #109, opened 2023-03-25
modified controls gui and made bringup gui & launch file

averystraumann:
added rosbag and dynamic reconfigure to controls rqt gui, and made bringup gui. copied all test-0x launch files into main launch directory in bringup so they can be accessed and launched directly from the bringup gui. bringup is untested as i cant run it on my machine


VasilyF:
Can you also delete the test-0x directories - so we don't have two versions of the same file

PR #111, opened 2023-03-28
made lane marker measure more robust

antoinedang:
+ able to detect straight lane markers as well as bent
+ most red pixel color, angle direction more reliable


VasilyF:
could we add the training set images in another repo? I don't want to run out of storage on the Jetson

antoinedang:
yeah sure I'll remove them

PR #126, opened 2023-05-31
Add DVL to quali.py

frompotenza:


antoinedang (changes requested):
the new interface uses kill() instead of 0'ing out the forces to make the AUV float back to the top

PR #135, opened 2023-06-10
Fix coordinates

MrMondrian:


frompotenza (approved):

frompotenza:
Might not work but maybe it works

PR #138, opened 2023-06-16
lane marker depth

frompotenza:


antoinedang (commented):
could be good to take a median or something but for now it should be fine

PR #143, opened 2023-06-20
Temp vision

frompotenza:


catkin_ws/src/vision/src/object_detection_utils.py
```diff
@@ -132,20 +132,37 @@ def clean_depth_error(depth_img, error_threshold=0.5):
...
-    return middle_point
+    middle_points = [depth_cropped[int(rows/2), int(cols/2)], depth_cropped[int(rows/2)+1, int(cols/2)],
+                    depth_cropped[int(rows/2)-1, int(cols/2)], depth_cropped[int(rows/2), int(cols/2)+1],
+                    depth_cropped[int(rows/2), int(cols/2)-1]]
+    return np.mean(middle_points)
 
 def lane_marker_depth(depth_cropped):
-    closest_point = np.unravel_index(np.argmin(depth_cropped), depth_cropped.shape)
-    return closest_point
+    """ Get 30 random points and take the mean - (some of them might be the floor but
```
antoinedang: I'm not sure random sampling is the best approach here, it's possible that the lane marker could represent 25% of the image, in which case the 30 points would be more likely to give the depth of the floor than of the lane marker itself. Might be better to filter out the background? I feel like it would be easier and better to filter out the background for all the objects, and then mean the remaining pixels?

catkin_ws/src/vision/src/object_detection_utils.py
```diff
@@ -246,8 +263,8 @@ def getObjectPosition(pixel_x, pixel_y, img_height, img_width, dist_from_camera=
 
 def measureBuoyAngle(depth_cropped):
     depth_cropped = clean_depth_error(depth_cropped, error_threshold=0.2)
-    rows, _ = depth_cropped.shape
-    left_half, right_half = depth_cropped[:rows/2, :], depth_cropped[rows/2:, :]
+    _, cols = depth_cropped.shape
+    left_half, right_half = depth_cropped[:, :int(cols/2)], depth_cropped[:, int(cols/2):]
```
antoinedang, line 273: This function (not related to this pull request but I just noticed) should give an exact angle, not just a 1 or -1. You can take the same math as elie did for the gate rotation (it's still in his branch) and I think it would pretty much the same, but would need to filter out the background first

antoinedang (changes requested):

catkin_ws/src/vision/src/object_detection_utils.py
frompotenza: what

catkin_ws/src/vision/src/object_detection_utils.py
MrMondrian: I think a high pass filter could work for some of the depth finding functions

antoinedang (approved):

PR #154, opened 2023-06-27
Added the octogone mission planner

Ayoubsab:


catkin_ws/src/planner/src/substates/octogone_task.py
```diff
@@ -0,0 +1,86 @@
...
+                print("No detection.... Timeout in " + str((startTime + timeout) - time.time()), end='\r')
+                if time.time() > startTime + timeout:
+                    print("Octogone no longer visible.")
+                    return 'failure'
+                    
+        print("Successfully centered octogone!")
+        return 'success'
+
+        
+class goThroughOctogone(smach.State):
```
frompotenza: name of the class should start with a Capital letter

catkin_ws/src/planner/src/mission.py
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env python3
 
+from planner.src.substates.octogone_task import CenterOctogone, goThroughOctogone
```
antoinedang: should be from src.substates.octogone_task import CenterOctogone, goThroughOctogone

catkin_ws/src/planner/src/mission.py
```diff
@@ -68,6 +69,22 @@ def Tricks(t):
             res = "trick not identified"
     endMission("Finished trick. Result {}".format(res))
 
+
+def octogone_task():
+    sm = smach.StateMachine(outcomes=['success', 'failure'])
+    ocotogne_target_class = 3 
+    with sm:
+        smach.StateMachine.add('find_octogone', BreadthFirstSearch(timeout=9999, target_classes=[(ocotogne_target_class, 1)], control=control, mapping=mapping),
```
antoinedang, line 79: We should move the AUV to the octagon approximate position before breadth first search

antoinedang: ideally should also support having several octagon position and at each failure we try the next one

catkin_ws/src/planner/src/substates/octogone_task.py
```diff
@@ -0,0 +1,86 @@
...
+        z_increment = 1
+
+        startTime = time.time()
+
+        while not (centered):
+            if len(last_object_detection) > 0:
+                startTime = time.time()
+                center_x = last_object_detection[0]
+                center_y = last_object_detection[1]
+                last_object_detection = []
```
antoinedang: doesn't use mapping, instead you should take mapping as an argument to this class, and use the methods supplied in src.substates.utility.mapping to navigate the gate -> you can assume if you've arrived at the navigation step that there is an octagon

catkin_ws/src/planner/src/substates/octogone_task.py
```diff
@@ -0,0 +1,86 @@
...
+
+                print("Octogone in view at: x:{}, y:{}!".format(center_x, center_y))
+                
+                centering_error = math.sqrt((center_x - targetCenterX)**2 + (center_y - targetCenterY)**2)
+
+                if centering_error > centering_tolerance: #isn't centered
```
antoinedang: once you use mapping you won't need this, you can just do control.move(octagon_x, octagon_y) to go above it

catkin_ws/src/planner/src/substates/octogone_task.py
```diff
@@ -0,0 +1,86 @@
...
+        
+    def execute(self, ud):
+        '''
+        Moves the AUV to the surface (ie, z=0)
+        '''
+        
+
+        print("Starting movement through octogone.")
+        curr_x, curr_y, _ = self.state.getPosition()
+        self.control.move((curr_x, curr_y, -1))
```
antoinedang: moving to your current position is redundant. You could just replace both the navigation and goThrough classes by this class and instead move to the octagon's x and y. the rest is correct

catkin_ws/src/planner/src/substates/octogone_task.py
```diff
@@ -0,0 +1,86 @@
...
+        self.control.kill()
+
+        return 'succes'
```
antoinedang: 'success'

antoinedang (changes requested):
see comments

PR #155, opened 2023-06-28
ez

frompotenza:


catkin_ws/src/auv_msgs/action/Quaternion.action
antoinedang, line 1: name Quaternion.action is not very clear, since it is both a quaternion and position action

catkin_ws/src/propulsion/src/thrust_mapper.py
```diff
@@ -19,27 +19,17 @@
...
 D_2 = 0.571 #m
 
 T = np.matrix(
-        [[-1., -1., 0., 0., 0., 0., 0., 0.],
-        [0., 0., 1., -1., 0., 0., 0., 0.],
-        [0., 0., 0., 0., -1., -1., -1., -1.],
-        [0., 0., 0., 0., -d/2, d/2, d/2, -d/2],
-        [0., 0., 0., 0., -D_2/2, -D_2/2, D_2/2, D_2/2],
-        [d/2, -d/2, hypot, hypot, 0., 0., 0., 0.]]
+        [[-1.,  -1.,  0.,    0.,    0.,     0.,     0.,     0.],
```
antoinedang, line 27: Make sure this is updated with recent changes to the thrust mapper matrix, when I merge it will revert those changes if they are on noetic

catkin_ws/src/auv_msgs/action/Quaternion.action
antoinedang: or it should not need to have do_x, do_y, etc.

catkin_ws/src/planner/src/substates/quaternion_test.py
antoinedang, line 1: unnecessary once we're merging it in, controller should inherently use quaternion server for its rotation (and position?)

antoinedang: the mission itself can stay but having a quaternion_action() function is not necessary

catkin_ws/src/planner/src/substates/utility/controller.py
```diff
@@ -4,7 +4,7 @@
 import actionlib
 from geometry_msgs.msg import Pose, Vector3, Vector3Stamped
 from std_msgs.msg import Float64, Bool, Header
-from auv_msgs.msg import StateAction, StateGoal, SuperimposerAction, SuperimposerGoal
+from auv_msgs.msg import StateAction, StateGoal, SuperimposerAction, SuperimposerGoal, QuaternionAction, QuaternionGoal
```
antoinedang, line 7: do we still need all these imports if we're just using the quaternion server?

catkin_ws/src/planner/src/substates/utility/controller.py
```diff
@@ -109,6 +114,26 @@ def get_state_goal(self,state,keepers,displace):
...
+        goal.pose.position.z = z
+        goal.pose.orientation.w = w
+        goal.pose.orientation.x = qx
+        goal.pose.orientation.y = qy
+        goal.pose.orientation.z = qz
+        goal.displace = displace
+        goal.do_x, goal.do_y, goal.do_z, goal.do_quaternion = keepers
+        return goal
+    
+    def quaternion_action(self,position,quaternion):
```
antoinedang, line 133: unnecesary function mentioned above

catkin_ws/src/planner/src/substates/utility/controller.py
```diff
@@ -109,6 +114,26 @@ def get_state_goal(self,state,keepers,displace):
...
+        goal = QuaternionGoal()
+        goal.pose.position.x = x
+        goal.pose.position.y = y
+        goal.pose.position.z = z
+        goal.pose.orientation.w = w
+        goal.pose.orientation.x = qx
+        goal.pose.orientation.y = qy
+        goal.pose.orientation.z = qz
+        goal.displace = displace
+        goal.do_x, goal.do_y, goal.do_z, goal.do_quaternion = keepers
```
antoinedang, line 130: still unclear whether or not quaternion server does position as well as quaternion rotation. Wouldn't it be simpler and easier to read if there were 3 servers: quaternion, position, and effort?

catkin_ws/src/planner/src/substates/utility/controller.py
```diff
@@ -18,6 +18,7 @@
 do_txtytz = [Bool(False),Bool(False),Bool(False),Bool(True),Bool(True),Bool(True)]
 do_tz = [Bool(False),Bool(False),Bool(False),Bool(False),Bool(False),Bool(True)]
 do_all = [Bool(True),Bool(True),Bool(True),Bool(True),Bool(True),Bool(True)]
+quaternion_dos = [Bool(True),Bool(True),Bool(True),Bool(True)]
 do_displace = Bool(True)
 do_not_displace = Bool(False)
```
antoinedang, line 24: having all of these definitions for different combinations of true/false seems overly complicated, why not 5 global variables: do_x, do_y, do_z, do_rot, do_displace?

catkin_ws/src/planner/src/mission.py
```diff
@@ -109,17 +118,5 @@ def master_planner():
...
-        QualiMission()
-
-
-        # ----- UNCOMMENT BELOW TO RUN MISSION(S) -----
-        #testRotationsMission()
-        #laneMarkerGridSearchMission()
-    except KeyboardInterrupt:
-        endMission("Mission end prompted by user. Killing.")
-        exit()
+    control = Controller(rospy.Time(0))
```
antoinedang: leave the try/except statement, it is useful to kill missions without having to constantly handle keyboard exceptions at different levels

catkin_ws/src/planner/src/substates/utility/controller.py
MrMondrian: We haven't fully transitioned from euler angles yet. Testing on clarke is necessary for us to deprecate the old code

catkin_ws/src/planner/src/substates/utility/controller.py
MrMondrian: Yes because right now we are only implementing code for experimenting with quaternions. Switching the implementation of move and rotate would be too much right now

catkin_ws/src/planner/src/substates/utility/controller.py
MrMondrian: For now we would still need a server that accommodates euler rotations until we can fully trust the quaternion code

catkin_ws/src/planner/src/substates/utility/controller.py
antoinedang: or nevermind I see now they're used to avoid defining them manually, but there has to be a better way. With superimposer + quaternion (I'm assuming we only have those 2), 6 DOFs for superimposer and 4 DOFs for quaternion, that's 6^2=36 options for superimposer, 4^2=16 options for quaternion, minus some because you can't have do_x on both at the same time, but you get my point

catkin_ws/src/planner/src/substates/utility/controller.py
```diff
@@ -47,6 +48,10 @@ def __init__(self, header_time):
         self.servers.append(self.SuperimposerServer)
         self.SuperimposerServer.wait_for_server()
 
+        self.QuaternionStateServer = actionlib.SimpleActionClient('quaternion_server', QuaternionAction)
```
antoinedang, line 51: quaternion_server is a misleading name for a server that does position + quaternion

antoinedang: if we're merging it in it should have the same name StateServer as before, what happens inside shouldn't be reflected to planner

catkin_ws/src/planner/src/substates/utility/controller.py
antoinedang: if Quaternion does x,y,z and rotation then we really need to A. integrate the goal sending to Quaternion server into the controller interface functions and B. remove any trace of the old server in planner (we can keep the class definition as a backup or make a backup of the current state of noetic to a branch but having 3 servers in the controller is confusing af)

If not then it shouldn't take in a Pose and we should make that clearer

catkin_ws/src/planner/src/substates/utility/controller.py
MrMondrian: Yeah its a bad name we can change it. But its unrealistic to think we're gonna replace the old stuff all at once. We need to merge temp code so we can test it before we can deprecate everything. I think we should merge it now and clean it up later. There isn't much point optimizing code that's still going to change

catkin_ws/src/planner/src/substates/utility/controller.py
antoinedang: we shouldn't have this intermediate state though, it's gonna be so messy and impossible to debug/test with

antoinedang: we should have this on a seperate branch and have no trace of euler angles on that branch, or save the euler angles version of noetic to another branch and remove any trace of them on noetic

MrMondrian: I don't see how thats the case. We only added one function to the planner that you can just ignore

catkin_ws/src/planner/src/substates/utility/controller.py
antoinedang: > 

I would hold off on merging then until we are sure it works at least to a certain extent. Gonna be so messy otherwise

antoinedang (changes requested):
I would only merge once we're 100% sure that this is the route we're taking and this branch has been cleaned up, it's more of a WIP rn and it would impede testing other things on noetic if we merged

catkin_ws/src/planner/src/substates/utility/controller.py
MrMondrian: We could do that but it would probably introduce like a million bugs

MrMondrian:
> I would only merge once we're 100% sure that this is the route we're taking and this branch has been cleaned up, it's more of a WIP rn and it would impede testing other things on noetic if we merged

Me and Felipe told you guys this branch is a work in progress but everyone agreed to move the code in anyway

catkin_ws/src/planner/src/substates/utility/controller.py
antoinedang: > 

The only advantage of merging right now that I see is to make it easier to test quaternions, and maybe during complex rotations (maybe?) avoiding Gimbal lock. The disadvantages are that we would have unstable and probably buggy code in the planner on the main branch that would make it much harder to test missions in the sim. I don't think it's a good idea to use the main branch as a temp merging branch, we should do that on a seperate branch no?

MrMondrian:
I can make the names better but switching everything over will introduce bugs. This PR won't because it doesn't touch anything that the planner uses

antoinedang:
> > I would only merge once we're 100% sure that this is the route we're taking and this branch has been cleaned up, it's more of a WIP rn and it would impede testing other things on noetic if we merged
> 
> Me and Felipe told you guys this branch is a work in progress but everyone agreed to move the code in anyway

Yeah I didn't realize it was this much a WIP, I really don't think it's a good idea to merge rn. Is there anything preventing us from having it seperate from noetic until it's ready?

MrMondrian:
No we can keep it separate

antoinedang:
> No we can keep it separate

ok bet let's do that imma close the PR if thats ok with u @MrMondrian

MrMondrian:
That's fine

PR #157, opened 2023-06-29
big cleanup, bug fixes, polishing, renaming for clarity, deleting red…

antoinedang:
PROBABLY USELESS SINCE QUATERNION PID WILL BE MERGED SOON


catkin_ws/src/auv_msgs/action/Superimposer.action
antoinedang, line 1: ignore it's just for global forces

catkin_ws/src/controls/launch/pid.launch
antoinedang, line 1: ignore just adding x and y back to pid

PR #162, opened 2023-07-02
correct thrust mapper based on measured thruster distance of new AUV …

VasilyF:
…and assumed thrust direction


MrMondrian:
Can you apply these changes to bridge.py in the sim too?

MrMondrian (approved):

VasilyF:
Yeah, I'm not sure this is correct, ideally I'd like to verify at the pool and then make the changes elsewhere

PR #167, opened 2023-07-04
Vision position timing fixes

antoinedang:
State in object detection is now paused when an image is received, and resumes updating when the image processing is finished. One state per thread (down cam and front cam).


frompotenza:
now that we have the models for down and front cam, would it be possible to test that in the sim before PRing?

PR #172, opened 2023-07-06
add reset_state_planar - to be used to orient the world frame to the …

VasilyF:
…aproximate orientation of UAV frame except keeping x-y axis planar with the global x,y axis (ie. z is always up)


catkin_ws/src/state_estimation/src/state_aggregator.py
```diff
@@ -64,9 +64,11 @@ def __init__(self):
         rospy.Subscriber("/sbg/ekf_quat", SbgEkfQuat, self.imu_cb)
         rospy.Subscriber("/dead_reckon_report",DeadReckonReport, self.dvl_cb)
         rospy.Subscriber("/depth", Float64, self.depth_sensor_cb)
-        rospy.Subscriber("/reset_state", Empty, self.reset_state_cb)
+        rospy.Subscriber("/reset_state_full", Empty, self.reset_state_full_cb)
+        rospy.Subscriber("/reset_state_planar", Empty, self.reset_state_planar_cb)
```
MrMondrian, line 68: what does planar mean. Is it supposed to be planner?

MrMondrian (approved):

catkin_ws/src/state_estimation/src/state_aggregator.py
VasilyF: it's alluding to keeping the the orientation of the x-y plane the same as the global frame (ie. the world frame is rotating within that plane if that makes sense). Maybe there's a better name for it?

PR #180, opened 2023-07-08
Goal displace

MrMondrian:
changes state quaternion server to displace goal instead of pose if the previous goal exists


MrMondrian:
I didn't realize @antoinedang already implemented this for rotations for today but my changes also implement this for position so I figured I'd PR all the changes anyway

catkin_ws/src/controls/src/servers/state_quaternion_pid_server.py
```diff
@@ -65,10 +69,17 @@ def callback(self, goal):
 
         self.server.set_succeeded()
 
-    def get_goal_after_displace(self):
-        goal_position = [self.pose.position.x + self.goal.pose.position.x, self.pose.position.y + self.goal.pose.position.y, self.pose.position.z + self.goal.pose.position.z]
-        if self.previous_goal_quat is None: self.previous_goal_quat = self.body_quat
-        goal_quat = self.previous_goal_quat * np.quaternion(self.goal.pose.orientation.w, self.goal.pose.orientation.x, self.goal.pose.orientation.y, self.goal.pose.orientation.z)
+    def get_goal_after_displace(self, goal):
+        if self.goal_position is None or self.goal_quat is None:
+            goal_position = [self.pose.position.x + goal.pose.position.x, self.pose.position.y + goal.pose.position.y, goal.pose.position.z + goal.pose.position.z]
```
VasilyF, line 74: what reference frame is the goal displacement relative to? I think this is fine if it's in the world frame same as self.pose (which I assume is taken from state_estimation)

However, if the displacement is relative to the AUV/body frame you need to account for the orientation of the AUV frame relative to the world frame (see state_aggregator update_pos_auv_world)

catkin_ws/src/controls/src/servers/state_quaternion_pid_server.py
```diff
@@ -65,10 +69,17 @@ def callback(self, goal):
...
+            goal_quat = self.body_quat * np.quaternion(goal.pose.orientation.w, goal.pose.orientation.x, goal.pose.orientation.y, goal.pose.orientation.z)
```
VasilyF, line 75: I believe the order of quaternion multiplication should be reversed (but double check). 

What I gather is: 
self.body_quat := q_auv_world (the quaternion rotating the world frame into the AUV frame)
goal.pose.orientation := q_goal_auv (quaternion rotating AUV frame to goal frame)

What you want is:
q_goal_auv*q_auv_world = q_goal_world (quaternion rotating world frame into goal frame)

catkin_ws/src/controls/src/servers/state_quaternion_pid_server.py
```diff
@@ -14,36 +14,40 @@ def __init__(self):
...
-            if(self.goal.displace.data):
-                goal_position, goal_quat = self.get_goal_after_displace()
+            if(goal.displace.data):
+                goal_position, goal_quat = self.get_goal_after_displace(goal)
+                self.goal_position = goal_position
+                self.goal_quat = goal_quat
             else:
-                goal_position = [self.goal.pose.position.x, self.goal.pose.position.y, self.goal.pose.position.z]
-                goal_quat = np.quaternion(self.goal.pose.orientation.w, self.goal.pose.orientation.x, self.goal.pose.orientation.y, self.goal.pose.orientation.z)
+                goal_position = [goal.pose.position.x, goal.pose.position.y, goal.pose.position.z]
```
VasilyF, line 30: The [StateQuaternionAction](https://github.com/mcgill-robotics/AUV-2023/blob/noetic/catkin_ws/src/auv_msgs/action/StateQuaternion.action) definition does contain a Pose - am I missing something? Later on you pass what I assume are meant to be booleans to the check_settled function. It seems you have a mix up of message data

catkin_ws/src/controls/src/servers/state_quaternion_pid_server.py
```diff
@@ -65,10 +69,17 @@ def callback(self, goal):
```
VasilyF, line 69: why sleep? instead of just letting the loop run?

VasilyF: I wouldn't let me comment on the actual line, it's the one above

catkin_ws/src/controls/src/servers/state_quaternion_pid_server.py
```diff
@@ -14,36 +14,40 @@ def __init__(self):
...
+            if(goal.displace.data):
+                goal_position, goal_quat = self.get_goal_after_displace(goal)
+                self.goal_position = goal_position
+                self.goal_quat = goal_quat
             else:
-                goal_position = [self.goal.pose.position.x, self.goal.pose.position.y, self.goal.pose.position.z]
-                goal_quat = np.quaternion(self.goal.pose.orientation.w, self.goal.pose.orientation.x, self.goal.pose.orientation.y, self.goal.pose.orientation.z)
+                goal_position = [goal.pose.position.x, goal.pose.position.y, goal.pose.position.z]
+                goal_quat = np.quaternion(goal.pose.orientation.w, goal.pose.orientation.x, goal.pose.orientation.y, goal.pose.orientation.z)
+                self.goal_position = goal_position
```
VasilyF, line 32: This is antoine from vasily's computer: this should be updated in the if statements, not for every goal. With this code, the goal positon/quaternion would always be updated to the most recent state goal, even if the do_x or d_y or dz or do_quat are false, in which case the goal position axes/quaternion should not be updated, since they will be 0 (instead of None). That's why I put the update for my implementation in the if goal.go_quat block so that it wouldn't be updated for position goals or wtv. Other than that LTGM

VasilyF (changes requested):
- position/quaternion previous goal update
- multiplication order for goal quaternion displacement

catkin_ws/src/controls/src/servers/state_quaternion_pid_server.py
MrMondrian: we can let it run if you want

PR #185, opened 2023-07-14
no_firmware launch of state_estimation to run state_aggregator alongs…

VasilyF:
…ide sim, thrust_mapper publish thruster_forces to use for bridge, this avoids having thrust matrix it two places


VasilyF:
these changes go along with https://github.com/mcgill-robotics/auv-sim/pull/30, before bridge was circumventing parts of state_estimation and propulsion which may lead to diverging behaviour

antoinedang (approved):

PR #199, opened 2023-08-03
bag all topics when running planner missions at transdec - allows eas…

VasilyF:
…y access to bag file on remote laptop


VasilyF:
run bringup on the jetson as normal. To bag topics run on laptop
`roslaunch planner testMission.launch bag:=true`

bag file should be in ~/.ros/

PR #206, opened 2023-09-01
Returned vision to using point cloud instead of RGB position estimation

antoinedang:
Also cleaned up some files


catkin_ws/src/planner/src/substates/navigate_buoy.py
```diff
@@ -47,23 +47,14 @@ def execute(self, ud):
...
-            if symbol_location == 1: # TOP LEFT
-                self.control.moveDeltaLocal(offset_distance, self.buoy_width/4, self.buoy_height/4)
-            elif symbol_location == 2: # TOP RIGHT
-                self.control.moveDeltaLocal(offset_distance, -self.buoy_width/4, self.buoy_height/4)
-            elif symbol_location == 3: # BOTTOM LEFT
-                self.control.moveDeltaLocal(offset_distance, self.buoy_width/4, -self.buoy_height/4)
-            elif symbol_location == 4: # BOTTOM RIGHT
-                self.control.moveDeltaLocal(offset_distance, -self.buoy_width/4, -self.buoy_height/4)
-
+            self.control.rotateEuler(homing_rotation) # bring to exact angle
```
MrMondrian, line 60: Very skeptical this will work. At the very least insert 0s into homing rotation instead of Nones

MrMondrian: In general this file has a lot of steps that aren't clear why they are there. For example the offset. Is that to put the AUV not too close?

MrMondrian (approved):
What's here will be a good starting point for the year. We should thoroughly review all of vision and mapping this fall.

catkin_ws/src/planner/src/substates/navigate_buoy.py
antoinedang: Yeah true, it could be cleaner for sure. I'll replace the None's with zeroes instead, but for the overall structure I feel like that's not a regression as much as something that was already a problem. We could give new members the task of cleaning up some of these missions/commenting them more thoroughly? Or I could do it now but I feel like lots of the missions are in a similar state

frompotenza:
are you able to test this in the sim? It looks good, but I think we should only start merging significant changes into the main branch once we can reliably run the missions in the sim without bypassing state estimation. Otherwise, we might end up with multiple problems at the same time and not be able to test anything properly.

MrMondrian:
We've never been able to run the sim without bypassing state estimation as far as I'm aware

frompotenza (approved):

catkin_ws/src/planner/src/substates/navigate_buoy.py
frompotenza: we could let the new members take care of this. It's mostly a copy-paste task with a little bit of ros formatting. It's good for making them familiar with how the planner works

antoinedang: Sounds good, I'll merge this in and add a task to the board for cleaning up missions/commenting

PR #243, opened 2023-09-09
Changed object detection frame message to be an array of object detections instead of arrays of fields (same for ObjectMap)

Ayoubsab:
## Related Issue(s)
  - Issue #219 

## Tests Done
- Not Tested


frompotenza:
render_visualization.py (vision) and vision.py (planner) were not updated

antoinedang (changes requested):
missing some updates in vision and planner files. It helps to do Ctrl+Shift+F in visual studio and search for any instances of ObjectDetectionFrame or ObjectMap. Other than that we can test it out and should be good to go!

catkin_ws/src/vision/src/object_detection.py
```diff
@@ -79,85 +84,93 @@ def detect_on_image(raw_img, camera_id):
...
+                    detectionFrame.extra_field = None
+
+                    detectionFrameArray.append(detectionFrame)
 
                     for symbol_class_name, symbol_x, symbol_y, symbol_z, confidence in buoy_symbols:
-                        label.append(symbol_class_name)
-                        obj_x.append(symbol_x)
-                        obj_y.append(symbol_y)
-                        obj_z.append(symbol_z) 
+                        detectionFrame = VisionObject()
```
antoinedang, line 150: this will overwrite the previously created instance of detectionFrame, but we want to add new detection Frames. So this should be renamed to something like symbolDetectionFrame

catkin_ws/src/vision/src/object_detection.py
```diff
@@ -79,85 +84,93 @@ def detect_on_image(raw_img, camera_id):
...
-                    obj_theta_z.append(None)
-                    extra_field.append(None)
+                        detectionFrame.theta_z = theta_z
+                        detectionFrame.extra_field = None
+
+                        detectionFrameArray.append(detectionFrame)
 
-    extra_field = [x if not x is None else -1234.5 for x in extra_field]
-    obj_theta_z = [x if not x is None else -1234.5 for x in obj_theta_z]
+
```
antoinedang, line 161: need to append detectionFrame to detectionFrameArray! otherwise we are only cleaning and sending the last detectionFrame that is created

catkin_ws/src/vision/src/object_detection.py
```diff
@@ -79,85 +84,93 @@ def detect_on_image(raw_img, camera_id):
...
-                    pred_obj_x, pred_obj_y, pred_obj_z = getObjectPositionFrontCam(bbox)
-                    obj_x.append(pred_obj_x)
-                    obj_y.append(pred_obj_y)
-                    obj_z.append(pred_obj_z) 
-                    obj_theta_z.append(None)
-                    extra_field.append(None)
+                        detectionFrame.theta_z = theta_z
+                        detectionFrame.extra_field = None
+
+                        detectionFrameArray.append(detectionFrame)
```
antoinedang, line 159: also this needs to be appended after the detectionFrame for the buoy, otherwise the order of the confidences will be messed up. Might make it easier if the confidence is an attribute of the VisionObject message and we can make sure they're always synced up. (sorry about that I only realized this looking at it now)

antoinedang (changes requested):
made some comments but other than that looks good, good job! Once the comments are addressed we can test it and make sure there's no glaring bugs and then merge it (vision isn't working perfectly right now so a quick test is fine)

catkin_ws/src/auv_msgs/msg/VisionObject.msg
```diff
@@ -0,0 +1,9 @@
+# Vision Object class
+string label
+#pose info
+float64 x
+float64 y
+float64 z
+float64 theta_z
+#additional data field (for additional info on objects like lane marker second heading)
+float64 extra_field
```
MrMondrian, line 9: I'm not a fan of this extra field thing here. Is this necessary?

antoinedang: it's since different objects have different features, so I just had extra_field there as a way of storing information which is object-specific (second heading for lane marker, which symbol is on the left for gate etc.). I agree it's probably not the cleanest solution but not part of the scope of this issue, if you want to change it we should make a seperate ticket and maybe give it to one of the new members since it'll be a relatively easy change

catkin_ws/src/vision/src/object_detection.py
Ayoubsab: I append the detection frame before entering the loop and I do it for each element in buoy_symbols

antoinedang: oh my bad, ignore this then

catkin_ws/src/vision/src/object_detection.py
Ayoubsab: I don't think it will mess anything because whenever I add a detectionFrame to the array I add the confidence too. So they are always in sync. 
But yeah adding the confidence as an attribute may be better. (It can be a good task for a new member)

antoinedang (approved):

catkin_ws/src/vision/src/object_detection.py
antoinedang: sorry about that you're right, good job👍

PR #260, opened 2023-09-20
Ayoub vision

Ayoubsab:
## Related Issue(s)
  - Issue #225 

## Changes Made
- Merged the function state, stateDelta, stateEuler, stateDeltaEuler in one function state which takes as arguments the position, the angle and two boolean isEuler and isDelta

## Tests Done
- No test done


catkin_ws/src/planner/src/substates/utility/controller.py
```diff
@@ -169,35 +169,35 @@ def rotateEuler(self,ang):
...
-        goal_state = self.get_state_goal([x,y,z,w,wx,wy,wz],do_not_displace)
-        self.StateQuaternionStateClient.send_goal_and_wait(goal_state)
-    
-    def stateDelta(self,pos,ang):
+
+    def state(self, pos, ang, isEuler=False, isDelta=False):
+        if isEuler:
+            wx,wy,wz = ang
+
+            if not isDelta:
```
MrMondrian, line 177: I don't think these two ifs need to be nested. could be `if isEuler and isDelta`

MrMondrian, line 177: when doing a delta control, you do that by passing do_displace to the function that gives you the goal. look at the other delta functions for inspiration. I don't think this if is necessary

MrMondrian (changes requested):

catkin_ws/src/planner/src/substates/utility/controller.py
antoinedang: I don’t know, I don’t see a better way of doing it. Could have the if statement included in the line but would be equivalent and would be slower. Maybe I’m missing something though

antoinedang (approved):

catkin_ws/src/planner/src/substates/utility/controller.py
antoinedang: @MrMondrian ?

MrMondrian: I think as a general programming practice the less nesting the better. This code doesn't need two ifs, one would do but how it would it be slower?

antoinedang: It does need two if's tho. If its a euler angle, convert it to quaternion. When you convert to quaternion, you need to know if its a delta or not to know what to replace None's with. So either way you put it you need two if's when there are eulers and quaternions combined. Maybe I'm missing something?

MrMondrian: yeah you're right my bad

MrMondrian:
Why are these changes requested in the first place? Seems more complex than necessary to be passing booleans for all the ways to interpret the data as opposed to having different functions. This also goes against the standard that the rotate functions follow. @antoinedang

antoinedang:
> Why are these changes requested in the first place? Seems more complex than necessary to be passing booleans for all the ways to interpret the data as opposed to having different functions. This also goes against the standard that the rotate functions follow. @antoinedang

I thought it would be cleaner to have state() with arguments rather than have state, stateDelta, stateEuler, stateDeltaEuler. Its true that having it different than rotation and move etc. is not very clean. I would argue tho that we should apply this pattern to those ones as well. having a single move, rotate, or state function instead of 10 different ones for different argument types makes it easier to maintain and read. Maybe we expand this issue into modifying rotate, move, and state all at once? We could even infer the type of rotation from the number of arguments. the function itself would be bigger but I like 3 big functions over 10 small ones. what do u think?

MrMondrian:
I think its easier to read mission planners with all the small functions as opposed to looking at all the arguments. I'm against having the function interpret the type of angle by the length of the array as that will probably cause bugs

MrMondrian:
While we're at it, now that flatten is implemented I think its also best to deprecate nones from the euler functions as those technically aren't mathematically correct and might be headache to change later if we have to

MrMondrian:
I also think its best if we force people to learn quaternions because if we give people this option that might not even work no one will learn quaternions. The only euler functions that I trust are rotateDeltaEuler, and rotateEuler without nones

antoinedang:
damn bruh I wrote a whole response and github bugged out and deleted my comment

antoinedang:
Let's talk about this tomorrow during the software meeting tomorrow, not tryna re-write everything. my main points were:
- dont think we should remove euler angles
- we can keep it seperate functions but in that case would be good to have a single function that handles the parsing/error checking of arguments to reduce code duplication and make it easier to maintain the control interface file
- im fine with removing None's for euler rotations

MrMondrian:
Isn’t the software meeting tonight lol. And we can for sure keep Euler angles we just need to be very cautious how we use them

antoinedang (approved):

Issue #241, opened 2023-09-09
Implement simple integration test using gazebo/github actions


MrMondrian:
Create an action that runs a simple test in sim

PR #273, opened 2023-09-27
Docker sim cherry pick

MrMondrian:
## Related Issue(s)
  - https://github.com/mcgill-robotics/AUV-2024/issues/241
  - https://github.com/mcgill-robotics/AUV-2024/issues/269

## Other Dependent PRs
  - https://github.com/mcgill-robotics/auv-sim-gazebo/pull/41

## Changes Made
- Adds rostest case for github actions that runs in gazebo

- Automates integration test
- Don't need to nag new members if they tested in sim
- Will be very easy to expand this for more tests

## Tests Done
- Tested locally using act, need to test on the main branch unfortunately

- Might be smart to add a page for all actions

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -5,7 +5,7 @@ on:
     branches: [noetic]
 
 jobs:
-  build:
+  CI:
     runs-on: ubuntu-20.04
     container:
         image: mcgillroboticsteam/auv_2024:main
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -1,4 +1,4 @@
-name: Catkin Build
+name: Integration Test
 
 on:
   pull_request:
@@ -14,6 +14,7 @@ jobs:
       - name: catkin build
         shell: bash
         run: |
+            git config --global --add safe.directory /__w/AUV-2024/AUV-2024
             git submodule update --init --recursive
             source /opt/ros/noetic/setup.bash
             cd catkin_ws
@@ -23,4 +24,4 @@ jobs:
         run: |
             cd catkin_ws
             source devel/setup.bash
-            rostest planner testMission.launch
+            rostest planner settle.test
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -14,6 +14,13 @@ jobs:
       - name: catkin build
         shell: bash
         run: |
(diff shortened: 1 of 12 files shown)


catkin_ws/src/planner/src/testMission.py
antoinedang, line 1: can we rename these files (test.launch, testMission.launch. testMission.py) into something clearer? Its no obvious from the name that the test is for continuous integration tests

catkin_ws/src/planner/src/testMission.py
```diff
@@ -2,11 +2,36 @@
...
 
-if __name__ == "__main__":
-    rospy.init_node("moveForward",log_level=rospy.DEBUG)
-    controls = Controller(rospy.Time(0))
-    controls.move((0,0,0))
+ 
+## A sample python unit test
+class Test1(unittest.TestCase):
+    ## test 1 == 1
+    def test__simple_movement(self): # only functions with 'test_'-prefix will be run!
```
antoinedang: while we're at it lets test some more complex orientations/position changes (can be done in a further PR, maybe lets make an issue out of it if we want to give this to a new member)

.github/workflows/pull_request.yml
```diff
@@ -14,6 +14,13 @@ jobs:
       - name: catkin build
         shell: bash
         run: |
+            git submodule update --init --recursive
             source /opt/ros/noetic/setup.bash
             cd catkin_ws
             catkin build
+      - name: simple test
```
antoinedang: can we rename the test from "simple test" to something like "sim test" or "planner test" or something so its clear when it fails what its testing

antoinedang (changes requested):
commented

antoinedang:
i forgot but also because of adding the sim as a git submodule we should update the setup instructions in the wiki

.github/workflows/pull_request.yml
MrMondrian: fixed

catkin_ws/src/planner/src/testMission.py
MrMondrian: lets save this for a new member

catkin_ws/src/planner/src/testMission.py
MrMondrian: done

antoinedang (approved):

frompotenza (approved):

Issue #295, opened 2023-10-04
Setup docker for ubuntu


MrMondrian:
Docker containers already work without guis. Just need to add graphics drivers and x11 forwarding

PR #308, opened 2023-10-07
Docker for ubuntu local development

MrMondrian:
## Related Issue(s)
  - https://github.com/mcgill-robotics/AUV-2024/issues/295

## Changes Made
- Add mesa drivers to dockerfile and main image to support visual rendering of ignition in docker. Only works in ubuntu. Add docker compose for easy launching of container that displays everything to the gui. Add optional gpu support for docker in the compose file that works

- Useful for if we get a PC in medn, we want to use our ubuntus for other things, and is a stepping stone towards being fully dockerized on all operating systems

## Tests Done
- Ran a simple mission fully dockerized with gui display
- Ran torch.cuda.is_available and it returned true

- May need to add setup instructions for ubuntu

- For gpu support, need to install nvidia-container-toolkit

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+text eol=lf
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,1 @@
-text eol=lf
+* text eol=lf
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,7 +1,9 @@
 [submodule "catkin_ws/src/auv-sim-gazebo"]
 	path = catkin_ws/src/auv-sim-gazebo
 	url = https://github.com/mcgill-robotics/auv-sim-gazebo.git
+	ignore = dirty
 
 [submodule "catkin_ws/src/auv-sim-unity"]
 	path = catkin_ws/src/auv-sim-unity
 	url = https://github.com/mcgill-robotics/auv-sim-unity.git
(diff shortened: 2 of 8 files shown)


frompotenza:
how smooth is the sim?

MrMondrian:
> how smooth is the sim?

Perfectly smooth on ubuntu. Preliminary tests on windows show it will be slow. I have confidence it will be smooth on mac

antoinedang (changes requested):
looks good, can you just make the necessary changes to install dependencies and the wiki? Then you can just bypass

MrMondrian:
> looks good, can you just make the necessary changes to install dependencies and the wiki? Then you can just bypass

The dependency would be installed outside the container so it wouldn't need to be in the install dependencies script

antoinedang:
> > looks good, can you just make the necessary changes to install dependencies and the wiki? Then you can just bypass
> 
> The dependency would be installed outside the container so it wouldn't need to be in the install dependencies script

Ok I guess just make the change to the wiki then, I forgot this would make it not necessary to run the install script lol

MrMondrian:
> Ok I guess just make the change to the wiki then, I forgot this would make it not necessary to run the install script lol

[Done](https://github.com/mcgill-robotics/AUV-2024/wiki/6.1-Docker)

Issue #223, opened 2023-09-05
Combine DVL + IMU + depth sensor packages into one Sensors package


antoinedang:

frompotenza:
Maybe put cameras in sensor package? Rename vision package to mapping?

PR #321, opened 2023-10-16
Single Sensors package for DVL + IMU + depth sensor

vivekandathil:
## Related Issue(s)
  - Issue #223 

## Changes Made
- Created a sensors/ package with combined CMakeLists.txt & package.xml file
- Moved related DVL + IMU + Depth Sensor files into the sensors package

## Tests Done
- Verified `catkin build` runs successfully (now with 15 packages instead of 17)


catkin_ws/src/sensors/depth_sensor/.placeholder
antoinedang, line 1: don't need this folder if it just has an empty file inside it

catkin_ws/src/sensors/CMakeLists.txt
```diff
@@ -0,0 +1,100 @@
...
+find_package(catkin REQUIRED COMPONENTS
+  roscpp
+  rosserial_arduino
+  rosserial_client
+  std_msgs  
+)
+
+catkin_package( CATKIN_DEPENDS
+  roscpp
+  rosserial_arduino
```
antoinedang, line 20: I think you're missing rospy in these dependencies (the imu has rospy in this part of the CMakeLists)

antoinedang: same for the find_package() part as well

catkin_ws/src/auv-sim-gazebo
MrMondrian, line 1: This submodule change would be breaking for docker on windows. @antoinedang I think we need a way to make the submodules always default to the latest commit on noetic.

For now this can be fixed by pulling the gazebo sim and commit the submodule changes again

MrMondrian:
Ignore the failing integration test it doesn't mean anything

antoinedang (changes requested):
Not sure I'm a fan of the seperate folders per-sensor, since they only have one (or zero) files in them. I think it would be cleaner to have a src/ and config/ folder in the top level with no subdirectories for sensors. Could you also make a README.md for the package (can be super short, just a brief description of what the package does)? Also, I think you forgot to update the imu.launch file (it looks for the imu config file using $(find imu)) which would cause an error. Last thing would be the CMakeLists.txt file is a bit of a mess with a lot of repeated parts, could you make sure any parts that are repeated are combined into one part (if possible) to make it more readable. Other than that good job and really quick too, gg

catkin_ws/src/auv-sim-gazebo
antoinedang: Yes, I agree I was about to say. I assumed that adding it to the gitignore would fix that stuff but apparently not

antoinedang:
damn you made those changes quick af lmao, good shit

antoinedang:
> Ignore the failing integration test it doesn't mean anything

It does @vivekandathil, you can ignore the CI if it fails the "settle" part of CI but for "catkin build" thats something to fix.

MrMondrian:
Oh I didn’t look to closely because it’s been failing on the settle test for some reason. The catkin build test is reliable

MrMondrian:
The integration test appears to be failing because the state estimation package is looking for the depth sensor package.

MrMondrian:
@vivekandathil the state estimation [launch file](https://github.com/mcgill-robotics/AUV-2024/blob/d040bcae486b1334f8bb2d87d1dd2a03fdd3acf9/catkin_ws/src/state_estimation/launch/state_estimation.launch#L4-L5) will also need to be modified as it looks for the dvl and imu launch files in the old packages.

catkin_ws/src/state_estimation/launch/state_estimation.launch
```diff
@@ -1,8 +1,8 @@
 <launch>
     <arg name="sim" default="false" />
     <group unless="$(arg sim)">
-		<include file="$(find imu)/launch/imu.launch"/>
-		<include file="$(find dvl)/launch/dvl.launch"/>
+		<include file="../../sensors/launch/imu.launch"/>
+		<include file="../../sensors/launch/dvl.launch"/>
```
MrMondrian: Would be better to do `$(find sensors)`

MrMondrian (changes requested):

catkin_ws/src/sensors/launch/imu.launch
```diff
@@ -1,5 +1,5 @@
 <launch>
 	<node name="sbg_device" pkg="sbg_driver" type="sbg_device" output="screen" respawn="true">
-		<rosparam command="load" file="$(find imu)/config/sbg_imu.yaml"/>
+		<rosparam command="load" file="../config/sbg_imu.yaml"/>
```
MrMondrian: this should be `$(find sensors)`

MrMondrian (changes requested):

catkin_ws/src/sensors/launch/dvl.launch
MrMondrian, line 1: I pulled the branch and tested state_estimation.launch. Line 6: `pkg="dvl"`. This causes a bug and needs to be updated to `pkg="sensors"`.

MrMondrian (changes requested):

catkin_ws/src/sensors/launch/dvl.launch
MrMondrian: I'm reasonably confident this will be the last change necessary

MrMondrian (approved):

catkin_ws/src/state_estimation/launch/state_estimation.launch
```diff
@@ -1,8 +1,8 @@
 <launch>
     <arg name="sim" default="false" />
     <group unless="$(arg sim)">
-		<include file="$(find imu)/launch/imu.launch"/>
-		<include file="$(find dvl)/launch/dvl.launch"/>
+		<include file="$(find sensors)/launch/imu.launch"/>
+		<include file="$(find sensors)/launch/dvl.launch"/>
```
antoinedang, line 6: I think this should also include the depth sensor launch file? (I see that it wasnt there before but that might have been a bug we hadnt found yet) Correct me if im wrong @MrMondrian

MrMondrian:
We probably commented it out at comp after the depth sensor broke

antoinedang (changes requested):
can you just add the depth sensor launch file to the state estimation launch file? After that all good

antoinedang (approved):

Issue #211, opened 2023-09-04
Test the visualization and calculation of the point cloud using the sim


antoinedang:

MrMondrian:
@antoinedang, wouldn't this be good for a new member?

PR #351, opened 2023-10-31
Fix bridge error in vision

vivekandathil:
## Related Issue(s)
  - Issue #211 

## Changes Made
- Removed lines causing bridge error when running the vision simulator

## Tests Done
- Verify catkin build
- Verify with simulator

### DIFF
--- a/catkin_ws/src/vision/src/object_detection_utils.py
+++ b/catkin_ws/src/vision/src/object_detection_utils.py
@@ -65,9 +65,9 @@ class State:
     def updatePose(self,msg):
         if self.paused: return
         self.q_auv = np.quaternion(msg.orientation.w, msg.orientation.x, msg.orientation.y, msg.orientation.z)
-    def updatePointCloud(self, msg):
+    def updatePointCloud(self):
         if self.paused: return
-        point_cloud_image = get_point_cloud_image(self.rgb, self.depth, self.y_over_z_map)
+        point_cloud_image = get_point_cloud_image(bridge, self.rgb, self.depth, self.width, self.height, self.x_over_z_map, self.y_over_z_map)
         if point_cloud_image is not None:
             self.point_cloud = np.copy(bridge.imgmsg_to_cv2(point_cloud_image, "passthrough"))
     def cleanPointCloud(self, point_cloud):
@@ -87,9 +87,12 @@ class State:
     def updateRGB(self, msg):
         temp = bridge.imgmsg_to_cv2(msg)
         self.rgb = temp/255
+        if self.depth is not None:
+            self.updatePointCloud()
     def updateDepth(self, msg):
         temp = bridge.imgmsg_to_cv2(msg)
         self.depth = temp/depth_scale_factor
+        self.updatePointCloud()
     def updateCameraInfo(self, msg):
         if(self.y_over_z_map is not None): return
         fx = msg.K[0]
@@ -104,6 +107,10 @@ class State:
 
         self.x_over_z_map = (cx - u_map) / fx
         self.y_over_z_map = (cy - v_map) / fy
+        if self.depth is not None:
+            self.updatePointCloud()
+        else:
+            print("Yo")
     def pause(self):
         self.paused = True
     def resume(self):
@@ -245,6 +252,7 @@ def measureAngle(bbox):
     zero_angle_vector = np.array([0,-1])
     arg_vector = right_avg_point - left_avg_point
     magnitude_arg_vector = np.linalg.norm(arg_vector)
+    # TODO: Check ValueError (dimensions don't match when doing dot product)
     dot_product = np.dot(zero_angle_vector, arg_vector)
     return math.acos(dot_product / magnitude_arg_vector) * 180 / math.pi
 
--- a/catkin_ws/src/vision/src/object_detection_utils.py
+++ b/catkin_ws/src/vision/src/object_detection_utils.py
@@ -109,8 +109,6 @@ class State:
         self.y_over_z_map = (cy - v_map) / fy
         if self.depth is not None:
             self.updatePointCloud()
-        else:
-            print("Yo")
     def pause(self):
         self.paused = True
     def resume(self):
--- a/catkin_ws/src/vision/src/object_detection_utils.py
+++ b/catkin_ws/src/vision/src/object_detection_utils.py
@@ -25,6 +25,12 @@ class State:
         self.paused = False
         self.q_auv = None
         self.point_cloud = None
+        self.depth = None
+        self.rgb = None
+        self.width = None
+        self.height = None
+        self.x_over_z_map = None
+        self.y_over_z_map = None
 
         self.x_pos_sub = rospy.Subscriber('state_x', Float64, self.updateX)
         self.y_pos_sub = rospy.Subscriber('state_y', Float64, self.updateY)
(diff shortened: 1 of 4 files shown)


MrMondrian (dismissed):

catkin_ws/src/vision/src/point_cloud.py
```diff
@@ -62,8 +62,8 @@ def convert_from_uvd(color, z_map):
         xyz_rbg_img[:, :, 0] = z_map
         xyz_rbg_img[:, :, 1] = x_map
         xyz_rbg_img[:, :, 2] = y_map
-        point_cloud_img = bridge.cv2_to_imgmsg(np.float32(xyz_rbg_img[:,:,:3]), "bgr8")
-        point_cloud_img_pub.publish(point_cloud_img)
+        # point_cloud_img = bridge.cv2_to_imgmsg(np.float32(xyz_rbg_img[:,:,:3]), "bgr8")
```
antoinedang: was this causing errors? I'm pretty sure this is important for the vision package, it uses this topic to get point cloud information. It should be replaced with listening directly to the point cloud topic and extracting rgb/xyz images directly but for now this should stay

antoinedang (changes requested):

MrMondrian:
This didn’t work on either of our machines and the code in point cloud .py is only intended for debugging

antoinedang:
> This didn’t work on either of our machines and the code in point cloud .py is only intended for debugging 

What didn't work about it? Cuz just deleting this line will break vision 100%. So I don't think we can merge this if it fixes one problem and causes another. Also I thought it was supposed to be a "real" file, I don't think having the point cloud math in object_detection_utils makes much sense. Or maybe we should rename it to like point_cloud_debug.py and have a point_cloud.py that handles the camera feeds and republishes a point cloud

MrMondrian:
The python implementation of the ros point cloud stuff is too slow. We can't transport point cloud data via ros for the time being. I profiled the function and the point cloud math is very fast even for large images but generating a point cloud ros message takes many seconds

MrMondrian:
When I wrote this I kept the vision point cloud and debugging point cloud separate. Maybe someone changed that but that would be a bad idea imo

catkin_ws/src/vision/src/point_cloud.py
MrMondrian: the bug can be fixed by removing the optional "bgr8" argument

antoinedang:
> When I wrote this I kept the vision point cloud and debugging point cloud separate. Maybe someone changed that but that would be a bad idea imo
> The python implementation of the ros point cloud stuff is too slow. We can't transport point cloud data via ros for the time being. I profiled the function and the point cloud math is very fast even for large images but generating a point cloud ros message takes many seconds

Ok makes sense, lets have object_detection_utils.py import point_cloud.py and use its functions, and then if we can have debugging be done by directly running the point_cloud.py

MrMondrian:
@antoinedang waiting on your approval

catkin_ws/src/vision/src/object_detection_utils.py
```diff
@@ -66,6 +70,7 @@ def cleanPointCloud(self, point_cloud):
         point_cloud[far_mask] = np.array([np.nan, np.nan, np.nan])
         return point_cloud
     def getPointCloud(self, bbox=None):
+        updatePointCloud()
```
antoinedang: should be self.updatePointCloud() I think

catkin_ws/src/vision/src/point_cloud.py
```diff
@@ -78,6 +63,27 @@ def convert_from_uvd(color, z_map):
...
+        point_cloud_img = bridge.cv2_to_imgmsg(np.float32(xyz_rbg_img[:,:,:3]), "bgr8")
+        return point_cloud_img
+
+def get_xyz_image(color, depth):
+    if y_over_z_map is not None:
+        xyz_rbg_img = np.zeros((height, width, 6))
+        xyz_rbg_img[:, :, 3:6] = color        
+
+        x_map = x_over_z_map * depth
+        y_map = y_over_z_map * depth
```
antoinedang: could you make the references to 'rbg' be 'rgb' (nitpick)

catkin_ws/src/vision/src/point_cloud.py
```diff
@@ -78,6 +63,27 @@ def convert_from_uvd(color, z_map):
...
         pub_msg = point_cloud2.create_cloud(header=header, fields=fields, points=xyz_rbg_img)
         return pub_msg
+
+def get_point_cloud_image():
+    if y_over_z_map is not None:
+        xyz_rbg_img = get_xyz_image(rgb, depth)
+        point_cloud_img = bridge.cv2_to_imgmsg(np.float32(xyz_rbg_img[:,:,:3]), "bgr8")
+        return point_cloud_img
+
+def get_xyz_image(color, depth):
```
antoinedang: since depth is a global variable no need for it to be passed as an argument I think

catkin_ws/src/vision/src/object_detection_utils.py
antoinedang: I don't think this works here for 2 reasons: since we never actually do any front cam detection until a point cloud exists (is not None), we'll never get to the point in the code where getPointCloud() is called. It'll be a sort of deadlock state where object_detection.py is waiting for the State class to have a non-None point cloud attribute, but the State class only has a non-None point cloud attribute when it's getPointCloud() function is called, and its only called after object_detection.py proceeds with a detection. Also, the same problem would happen with the pausing: getPointCloud() should only be called after the State class is paused, which means that the point cloud will never actually update since it's always paused

antoinedang: I think it might be cleaner and avoid these problems by creating a subscriber inside the State class for image and camera_info topics and using the functions from point_cloud.py to turn them into a usable point cloud array. It would also avoid having to use global variables for depth

antoinedang (changes requested):
good code and I like the depth_scaling_factor changing based on sim vs. real, just re-structuring would fix it

catkin_ws/src/vision/src/object_detection_utils.py
```diff
@@ -31,7 +39,11 @@ def __init__(self):
         self.theta_x_sub = rospy.Subscriber('state_theta_x', Float64, self.updateThetaX)
         self.theta_y_sub = rospy.Subscriber('state_theta_y', Float64, self.updateThetaY)
         self.theta_z_sub = rospy.Subscriber('state_theta_z', Float64, self.updateThetaZ)
-        self.point_cloud_sub = rospy.Subscriber('vision/front_cam/point_cloud_raw', Image, self.updatePointCloud, queue_size=1)
+        # Update the point cloud whenever the current image is updated
+        self.camera_info_sub = rospy.Subscriber('/vision/front_cam/aligned_depth_to_color/camera_info', CameraInfo, self.updateCameraInfo)
+        self.depth_sub = rospy.Subscriber('/vision/front_cam/aligned_depth_to_color/image_raw', Image, self.updateDepth)
+        self.rgb_sub = rospy.Subscriber('/vision/front_cam/color/image_raw', Image, self.updateRGB)
```
antoinedang, line 45: we don't need an RGB subscriber normally

catkin_ws/src/vision/src/object_detection_utils.py
```diff
@@ -70,6 +84,33 @@ def getPointCloud(self, bbox=None):
...
+
+        u_map = np.tile(np.arange(self.width),(self.height,1)) +1
+        v_map = np.tile(np.arange(self.height),(self.width,1)).T +1
+
+        self.x_over_z_map = (cx - u_map) / fx
+        self.y_over_z_map = (cy - v_map) / fy
+        if self.depth is not None:
+            self.updatePointCloud()
+        else:
+            print("Yo")
```
antoinedang: lmao what does "Yo" mean?

antoinedang (dismissed):
since its been a while and you guys will be fixing this stuff anyway, I'm good to merge but you'll have to make sure in the future that we can clean this up (just the rgb subscriber stuff and commenting, handling the value error ,etc.). I'll leave it up to you

antoinedang:
Oh also make sure the integration test doesnt fail before you merge

catkin_ws/src/vision/src/object_detection_utils.py
vivekandathil: Ah sorry about that! Forgot to remove after testing haha

antoinedang:
@vivekandathil good to merge for me, especially since it'll unblock everyone working on front cam stuff. Btw feel free to resolve any conversations you've addressed before, don't wanna do it for you in case you are still addressing any but yeah i dont mind if you do

vivekandathil:
Sounds good, thanks! @antoinedang 
I'll create tasks for fixing the RGB subscriber & ValueError parts after merging

Issue #266, opened 2023-09-26
Change sensor priority so that the dvl depth always picks up exactly where the depth sensor left off


MrMondrian:
Currently if the depth sensor goes offline, then when the dvl depth takes over it will jump to whatever the dvl thinks the depth is. Instead an offset should be applied to prevent the jump from occuring

antoinedang:
NOTE: we could also apply something similar to DVL/IMU (I think going from IMU -> DVL should be good already because it uses the IMU's reference frame to correct, but DVL->IMU will jump). Correct me if im wrong @MrMondrian

MrMondrian:
You're right, but is this case it might be better to let it jump because the IMU's quaternion is what we trust more. I'm imagining things going really weirdly if they flip flop multiple times in a row. Letting it jump could be dangerous though. @antoinedang

antoinedang:
> You're right, but is this case it might be better to let it jump because the IMU's quaternion is what we trust more. I'm imagining things going really weirdly if they flip flop multiple times in a row. Letting it jump could be dangerous though. @antoinedang

That's true, let's keep it to just depth then. If it becomes a problem that they jump we could implement some sort of damping/interpolation between the two (would probably be over-engineering it tho lol). For now I agree lets not fix that problem unless it becomes one and just do depth

PR #353, opened 2023-11-04
change sensor priority

tarekelakkaoui:
## Related Issue(s)
  - Issue #266 

## Other Dependent PRs
  - None

## Changes Made
- The z value taken is the active sensor's z value and the previous working sensor's z value.
- My reasoning is to take the value of z value that is being read and shift its value by the previous sensor's value.

## Tests Done
- Catkin build
- None

- None
- None

### DIFF
--- a/catkin_ws/src/state_estimation/src/state_aggregator.py
+++ b/catkin_ws/src/state_estimation/src/state_aggregator.py
@@ -33,20 +33,20 @@ def update_state(_):
             break
     for sensor in sensor_priorities["z"]:
         if sensor.isActive():
-        	if sensor.sensor_name == "depth_sensor":
-        		ds_z = sensor.z
-        		if dvl_z is not None:
+            if sensor.sensor_name == "depth_sensor":
+                ds_z = sensor.z
+                if dvl_z is not None:
                     offset = math.abs(ds_z - dvl_z)
                     i += 1
-        			ds_z -= math.exp(-i)*offset
-        		z = ds_z
-        	else:
-        		dvl_z = sensor.z
-        		if ds_z is not None:
+                    ds_z -= math.exp(-i)*offset
+                z = ds_z
+            else:
+                dvl_z = sensor.z
+                if ds_z is not None:
                     offset = math.abs(ds_z - dvl_z)
-        			dvl_z -= offset
-        		z = dvl_z
-        	break
+                    dvl_z -= offset
+                z = dvl_z
+                break
     for sensor in sensor_priorities["orientation"]:
         if sensor.isActive():
             np_quaternion = np.array([sensor.q_nwu_auv.x, sensor.q_nwu_auv.y, sensor.q_nwu_auv.z, sensor.q_nwu_auv.w])
--- a/catkin_ws/src/state_estimation/src/state_aggregator.py
+++ b/catkin_ws/src/state_estimation/src/state_aggregator.py
@@ -36,14 +36,14 @@ def update_state(_):
         	if sensor.sensor_name == "depth_sensor":
         		ds_z = sensor.z
         		if dvl_z is not None:
-                    		offset = math.abs(ds_z - dvl_z)
-                    		i += 1
+                    offset = math.abs(ds_z - dvl_z)
+                    i += 1
         			ds_z -= math.exp(-i)*offset
         		z = ds_z
         	else:
         		dvl_z = sensor.z
         		if ds_z is not None:
-                    		offset = math.abs(ds_z - dvl_z)
+                    offset = math.abs(ds_z - dvl_z)
         			dvl_z -= offset
         		z = dvl_z
         	break
--- a/catkin_ws/src/state_estimation/src/state_aggregator.py
+++ b/catkin_ws/src/state_estimation/src/state_aggregator.py
@@ -5,7 +5,7 @@ import numpy as np
 import quaternion
 from tf import transformations
 from tf2_ros import TransformBroadcaster
-from math
+import math
 
 from geometry_msgs.msg import Pose, Quaternion, Vector3, TransformStamped, Point
 from sensors import DepthSensor, IMU, DVL
@@ -36,14 +36,14 @@ def update_state(_):
         	if sensor.sensor_name == "depth_sensor":
         		ds_z = sensor.z
         		if dvl_z is not None:
-                    offset = math.abs(ds_z - dvl_z)
-                    i += 1
+                    		offset = math.abs(ds_z - dvl_z)
+                    		i += 1
         			ds_z -= math.exp(-i)*offset
         		z = ds_z
         	else:
         		dvl_z = sensor.z
         		if ds_z is not None:
-                    offset = math.abs(ds_z - dvl_z)
+                    		offset = math.abs(ds_z - dvl_z)
         			dvl_z -= offset
         		z = dvl_z
         	break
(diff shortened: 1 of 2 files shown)


catkin_ws/src/state_estimation/src/state_aggregator.py
```diff
@@ -29,8 +31,17 @@ def update_state(_):
             break
     for sensor in sensor_priorities["z"]:
         if sensor.isActive():
-            z = sensor.z
-            break
+        	if sensor.sensor_name == "depth_sensor":
+        		ds_z = sensor.z
```
antoinedang: I think this wouldn't function as expected. We know that the depth sensor is always the more accurate one, so we need a solution that 1. when depth sensor is available, saves this value 2. when only the DVL is available, applies the difference between the DVL reading and the depth sensor reading as a constant so that the dvl reading will continue from where the depth sensor left off, and save the value of the latest DVL reading (including offset) 3. when the depth sensor turns back on, we want to use the latest DVL reading (including offset) to make sure the depth sensor continues from where the DVL left off, but at each new reading we slowly decrease the offset. The behaviour would then be that the transition between the DVL and depth sensor readings is smooth but we also want to converge to where the depth sensor says we are since its the most accurate reading. Let me know if that makes sense! You're close with this implementation so good job, and you correctly found where to make the changes. Just need to switch around a few variables and define an interpolation method for when the depth sensor turns back on and we're good!

antoinedang (changes requested):

MrMondrian:
@tarekelakkaoui the integration test failed it looks like there is a bug. you can check the logs here https://github.com/mcgill-robotics/AUV-2024/actions/runs/6792732608/job/18466481504?pr=353

tarekelakkaoui:
I'm not really sure why my indentation is not formatting correctly.

MrMondrian:
> I'm not really sure why my indentation is not formatting correctly.

sometimes python doesn't like it if you're editing with vim or nano

catkin_ws/src/state_estimation/src/state_aggregator.py
```diff
@@ -29,8 +33,20 @@ def update_state(_):
...
+                ds_z = sensor.z
+                if dvl_z is not None:
+                    offset = math.abs(ds_z - dvl_z)
+                    i += 1
+                    ds_z -= math.exp(-i)*offset
+                z = ds_z
+            else:
+                dvl_z = sensor.z
+                if ds_z is not None:
+                    offset = math.abs(ds_z - dvl_z)
```
antoinedang: I think this wouldnt work again. Here you're taking the absolute difference between the last depth sensor reading and the current dvl reading, and applying that offset to the DVL. The problem with that is that 1. the DVL reading will always be equal to the depth sensor, since the offset that is calculated is not saved and re-used, so essentially the DVL is always being offset by an amount so it is equal to the last depth sensor value. What we want is that the first time the DVL is available and the depth sensor is not, we save the offset, and then on later iterations we always apply that offset that was calculated the first time. Also, using an absolute value doesnt work, when you subtract an absolute value from the DVL's reading you're making the assumption that the DVL's reading will always be greater than the depth sensor's last reading, which isn't true.

catkin_ws/src/state_estimation/src/state_aggregator.py
```diff
@@ -29,8 +33,20 @@ def update_state(_):
...
+                    dvl_z -= offset
+                z = dvl_z
```
antoinedang: Could you also include comments for what is going on here and why? I think it would make it easier for you to debug and also for me to review

catkin_ws/src/state_estimation/src/state_aggregator.py
```diff
@@ -29,8 +33,20 @@ def update_state(_):
...
     for sensor in sensor_priorities["z"]:
         if sensor.isActive():
-            z = sensor.z
-            break
+            if sensor.sensor_name == "depth_sensor":
+                ds_z = sensor.z
+                if dvl_z is not None:
+                    offset = math.abs(ds_z - dvl_z)
+                    i += 1
+                    ds_z -= math.exp(-i)*offset
```
antoinedang: choice of exp() is good, just make sure to remember to reset i to 0 when the depth sensor turns off! Also same problem here as with the DVL with the offset always being re-calculated

antoinedang (changes requested):

antoinedang:
Also dont get discouraged by all the comments, I'm being nitpicky but this solution is really close to being correct, and the problem with the tabs/spaces is not a big deal

catkin_ws/src/state_estimation/src/state_aggregator.py
```diff
@@ -29,8 +33,34 @@ def update_state(_):
...
+                    if dvl_z is not None: #if switch has occurred
+                        # when i increases, exp decreases and converges to 0
+                        i += 1
+                        ds_z -= math.exp(-i)*offset
+                    z = ds_z
+                    break
+                else: # dvl
+                    # re-calculate offset for each switch
+                    dvl_z = sensor.z
+                    if i != 0:
```
antoinedang, line 56: I think `i` needs to be initialized to a non-zero value (if the depth sensor is active from the start the dvl should grab an offset even if the depth sensor has not transitioned yet)

catkin_ws/src/state_estimation/src/state_aggregator.py
```diff
@@ -29,8 +33,34 @@ def update_state(_):
...
+                        #generates the offset
+                        offset = dvl_z - ds_z
+                    if ds_z is not None: # if ds functional from the start
```
antoinedang, line 59: here you check if the depth sensor value is None after using it already (in the if statement above), so you should have an outer if statement "if ds_z is not None" and then inside that check if i is nonzero

antoinedang (changes requested):
like 2 lines are off, but other than that good shit

catkin_ws/src/state_estimation/src/state_aggregator.py
tarekelakkaoui: So should i just initialize i to (for example) 1, and then keep the same condition and the reset to 0?

MrMondrian:
Honestly I'm having a hard time reviewing this PR. @tarekelakkaoui can you test this in sim? You can mimic the depth sensor cutting out by modifying auv-sim-gazebo/src/bridge.py where it sends the depth sensor data. You can use something like np.random.rand and rospy.get_time so randomly stop sending depth sensor data for certain periods of time

@antoinedang would you agree this is the right approach? I don't want this PR going stale but its hard to figure out if its bug free just by looking at it, and I think I see some bugs but I'm not 100%

MrMondrian:
Also when did we decide to do the interpolation thing? I think its overkill. We should get this working with just the offset first then add interpolation once that works because its @tarekelakkaoui first issue

antoinedang:
> Also when did we decide to do the interpolation thing? I think its overkill. We should get this working with just the offset first then add interpolation once that works because its @tarekelakkaoui first issue

Yeah it wasn't originally planned but the whole task doesn't make sense without it. When we go depth sensor -> DVL for z position we want the transition to be smooth (i.e DVL starts from where depth sensor left off). If when we go DVL -> depth sensor there's a jump the problem hasnt't been fixed. But since the depth sensor is more accurate we need the readings to smoothly converge to where the depth sensor says the z position is, not just continue off from where the DVL left off. So it became necessary even though wasnt initially specified

antoinedang:
Also @tarekelakkaoui was going to finish the PR I think? It's very close to done, so maybe we should keep it open until he wants to work on it again. (or I can finish it if its stays open too long) @MrMondrian

MrMondrian:
I told him not to finish it because its low priority and stale so it will conflict with the current higer priority state estimation fixes stuff

PR #386, opened 2023-11-15
Integrate dry-test package into the propulsion package (no need for it to be its own package)

NA29:
Moved the launch and src files from dry-test to propulsion and deleted dry-test folder.


frompotenza:
Make sure to add this to the wiki

frompotenza:
@MrMondrian do you prefer to merge this after we dry test? the problem is that the auv is at our place

MrMondrian:
Yeah we can merge this after we test it on the auv

PR #403, opened 2023-11-22
Buoyant Force Offset

epicgamer17:


antoinedang (dismissed):

catkin_ws/src/controls/src/superimposer.py
```diff
@@ -52,9 +52,16 @@ def update_effort(self, _):
         roll = self.roll.val
         pitch = self.pitch.val
         yaw = self.yaw.val
+        
+	#Toggled in state_quaternion_pid_server.py
+        if (rospy.get_param("enable_buoyant_force_offset")):
```
MrMondrian: I think the communication should be done through publishers and subscribers, not through params. The param you would set once at the start to whether or not the bouyant force offset should ever be used, then the controls server and superimposer would communicate through a topic whether or not the bouyant offset force should be active at this given moment. The reason this is better is so that in sim we might not want this behaviour, especially in the gazebo sim where the is no "water level" so it doesn't make sense for the code to change controls close to the "surface".

MrMondrian (changes requested):

catkin_ws/src/controls/src/superimposer.py
antoinedang: I see your point, not sure the solution is necessarily publisher/subscribers, we could just add a param that globally disables/enables buoyant forces and use this param to do live toggles of whether we apply it or not

antoinedang: I feel like adding a topic just for enabling/disabling a variable would be more complexity in the code for the same effect as a param

MrMondrian: the problem with this code is that it prevents us from using the param. If we set it to be off before running then that value gets overwritten as soon as a new goal is received. This type of communication is kinda the whole point of the pub/sub system

antoinedang: yeah no I see what you're saying, we can add a topic to fix that, I was just saying it could be a topic or a param, same effect in the end

PR #407, opened 2023-11-26
Felps screen display

frompotenza:
## Related Issue(s)
  - Code for display screen

## Changes Made
- Embedded code for display (+changing some cmake and launch stuff - nothing major)
- Add publisher for sensor status in state estimation and for mission in planner

## Tests Done
- Test for mission status: Run sim for different missions and check the display
- Test for sensor status: Stop publishing to the sensor topic (remove publisher from bridge.py - sim)


MrMondrian:
What is the semi colon commit in the middle?

catkin_ws/src/sensors/launch/display.launch
```diff
@@ -0,0 +1,5 @@
+<launch>
+    <node name="display_serial_server" pkg="rosserial_python" type="serial_node.py" respawn="true">
+        <param name="port" value="/dev/ttyACM0"/>
```
MrMondrian, line 3: This port will need to change at some point with udev rules but this is fine for now

catkin_ws/src/state_estimation/src/sensors.py
MrMondrian, line 1: what do the changes in this file do? I see a bunch of msg variables that aren't used

catkin_ws/src/sensors/embedded/CMakeLists.txt
MrMondrian, line 1: Are these changes to have in case we want to switch to flashing with catkin?

MrMondrian (commented):
Looks great just have a couple questions

catkin_ws/src/sensors/embedded/CMakeLists.txt
frompotenza: Idk if this actually works. I tried to copy-paste some stuff, but I didn't really need it

frompotenza:
> What is the semi colon commit in the middle?

I think that was me pulling from noetic

MrMondrian (approved):

PR #421, opened 2023-12-07
add test certificate verification on new commit

antoinedang:
## Related Issue(s)
[issue](https://github.com/mcgill-robotics/AUV-2024/issues/420)

## Changes Made
- Replaced github action hosted integration tests with a local integration test suite and verification of the testing on github actions.

## Tests Done
- this PR will test

- workflow section


MrMondrian:
did you add the encrypted bash scripts so that they can be edited by future members? I think its fine just to add the binary. but asking members to run random binaries is lowkey kinda sus tho. also the binary won't work on arm computers

antoinedang:
> did you add the encrypted bash scripts so that they can be edited by future members? I think its fine just to add the binary. but asking members to run random binaries is lowkey kinda sus tho. also the binary won't work on arm computers

yeah i kept the encrypted bash script so it can be iterated. If i just keep the binary then the script is lost and if we update any of the tests or whatever or want to add selective testing it would have to be re-written from scratch, so I feel like it makes sense to keep them there in case. I can put them somewhere else if necessary

Good point for the ARM thing, didnt know about that. Seems like from this post https://github.com/neurobin/shc/issues/33 its possible to compile to ARM, so I'll do that now just need someone with ARM to test it.

MrMondrian:
I think what you have here works but wouldn't it make more sense to try and use manually triggered actions to reduce minutes instead of locally running? I think maybe this is more of a last resort if we manage to blow through all our minutes even with manually running. [This post](https://stackoverflow.com/questions/63288356/how-to-link-a-github-actions-manual-run-to-a-pr) explains how to trigger pr test from manually triggered workflows

@frompotenza what are your thoughts

antoinedang:
> I think what you have here works but wouldn't it make more sense to try and use manually triggered actions to reduce minutes instead of locally running? I think maybe this is more of a last resort if we manage to blow through all our minutes even with manually running. [This post](https://stackoverflow.com/questions/63288356/how-to-link-a-github-actions-manual-run-to-a-pr) explains how to trigger pr test from manually triggered workflows
> 
> @frompotenza what are your thoughts

Yeah no for sure, it would be good to have both as an option. Like either test it manually on github actions w/ a comment or run it locally. Do you want me to add that feature inside this PR or in a seperate one? I assumed it would be seperate but at once makes sense too

MrMondrian:
I think a separate PR is best. I would make a manually triggered workflow that takes in the PR number as input, fetches the data with the github api, gets the commit, checks it out, does the test, then uses this action to set the status https://github.com/myrotvorets/set-commit-status-action

antoinedang:
> I think a separate PR is best. I would make a manually triggered workflow that takes in the PR number as input, fetches the data with the github api, gets the commit, checks it out, does the test, then uses this action to set the status https://github.com/myrotvorets/set-commit-status-action

sounds good, I'll add it to the board. Good to merge though?

antoinedang:
> > I think a separate PR is best. I would make a manually triggered workflow that takes in the PR number as input, fetches the data with the github api, gets the commit, checks it out, does the test, then uses this action to set the status https://github.com/myrotvorets/set-commit-status-action
> 
> sounds good, I'll add it to the board. Good to merge though?

btw integration test failed cuz i merged noetic in so the test certificate is invalid now. I looked into checking the contents of the catkin_ws/src folder instead of using commit hash but generating a certificate with those contents took super long. Would definitely be better though for the future

MrMondrian:
> sounds good, I'll add it to the board. Good to merge though?

Well if we add this we will have to enforce it and if it ends up being a challenge for new members it could hold us back. I would like to verify it works on arm first

antoinedang:
closing this because it won't apply to the new sim, we can re-visit it later on if necessary

PR #422, opened 2023-12-07
Fix Value Error when running state estimation

antoinedang:
fix was straightforward: subscriber was initialized before .imu attribute of class which would cause a ValueError if the subscriber callback was called before the init function of a sensor class was finished


MrMondrian:
Okay sounds good I'm killing the workflow because we don't need to waste 40 mins of actions for this lol

MrMondrian (approved):

antoinedang:
> Okay sounds good I'm killing the workflow because we don't need to waste 40 mins of actions for this lol

fax lmao

PR #431, opened 2023-12-10
State estimation fixes

MrMondrian:
## Related Issue(s)
  - Issue #1
  - Issue #2

## Changes Made
- Fixes annoying bugs in sensors.py/state_aggregator.py. Ex: imu says active when it gets an angular velocity message before a quaternion message resulting in an access to imu.q_nwu_auv when it is None. sensors had roll pitch yaw values that weren't being used, state_aggregator was doing the conversion from the quaternion and publishing that. Changes update state to force all sensor subclasses to implement the checks for valid state and different state themselves which significantly cleans the code. Comment out parts of the code that we are not confident about. Remove dvl from orientation list temporarily

## Tests Done
- Tested with physical imu
- Ran all integration tests locally


antoinedang:
lots of silly bugs i see

PR #446, opened 2024-01-09
change imu offset method

MrMondrian:


catkin_ws/src/state_estimation/launch/state_estimation.launch
```diff
@@ -6,10 +6,10 @@
 		<include file="$(find sensors)/launch/dvl.launch"/>
     </group>
 
-	<arg name="q_imu_auv_w" default="0.0" />
-	<arg name="q_imu_auv_x" default="0.0" />
-	<arg name="q_imu_auv_y" default="0.0" />
```
antoinedang, line 11: think the way its currently mounted is 180 on X, not important though. Good to merge for me

antoinedang (dismissed):

catkin_ws/src/sim/launch/sim.launch
```diff
@@ -18,23 +18,22 @@
     <arg name="vision" default="false" />
 
     <param name="sim" value="true" />
-
+    <param name="q_imunominal_imu_w" value="1.0" />
+    <param name="q_imunominal_imu_x" value="0.0" />
+    <param name="q_imunominal_imu_y" value="0.0" />
+    <param name="q_imunominal_imu_z" value="0.0"/>
+
+    <param name="q_dvl_auv_w" value="0.0" />
```
antoinedang, line 26: does this work? wouldn't the state estimation param values overwrite these ones if it's launched after

MrMondrian: no state estimation only does that if the sim arg is false

antoinedang: oh ok i see thats gucci then

catkin_ws/src/state_estimation/launch/state_estimation.launch
```diff
@@ -4,40 +4,22 @@
...
-	<arg name="auv_dvl_offset_y" default="0.0" />
-	<arg name="auv_dvl_offset_z" default="-0.3" />
 
-	<node name="state_aggregator" pkg="state_estimation" type="state_aggregator.py" respawn="true" output="screen">
-		<param name="update_rate" value="100" />
-		<param name="q_imu_auv_w" value="$(arg q_imu_auv_w)" />
-		<param name="q_imu_auv_x" value="$(arg q_imu_auv_x)" />
-		<param name="q_imu_auv_y" value="$(arg q_imu_auv_y)" />
-		<param name="q_imu_auv_z" value="$(arg q_imu_auv_z)" />
+		<param name="q_imunominal_imu_w" value="1.0" />
```
antoinedang, line 8: 180 on X for the IMU currently

PR #455, opened 2024-01-11
add hydrophones

frompotenza:
Add hydrophones to the code. It is not in a catkin package, so it's not going to cause any problems. I'm PRing early because the sim already has the hydrophones msgs. So, if anyone tries to test ros stuff in the sim, they'll get an error

### DIFF
--- a/catkin_ws/src/auv_msgs/msg/UnityState.msg
+++ b/catkin_ws/src/auv_msgs/msg/UnityState.msg
@@ -2,6 +2,8 @@ geometry_msgs/Vector3 position
 geometry_msgs/Quaternion orientation
 geometry_msgs/Vector3 velocity
 geometry_msgs/Vector3 angular_velocity
+geometry_msgs/Vector3 pinger1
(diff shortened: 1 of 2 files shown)


MrMondrian:
is that @Rain1618's code or legacy code

MrMondrian (approved):

frompotenza:
> is that @Rain1618's code or legacy code

it's her code

catkin_ws/src/auv_msgs/msg/UnityState.msg
```diff
@@ -2,6 +2,8 @@ geometry_msgs/Vector3 position
 geometry_msgs/Quaternion orientation
 geometry_msgs/Vector3 velocity
 geometry_msgs/Vector3 angular_velocity
+geometry_msgs/Vector3 pinger1
```
antoinedang, line 5: can we rename to something clearer? like dt_pinger1 or dt_20kHz or some shit

antoinedang (changes requested):

PR #457, opened 2024-01-13
Vision classes

### LINKED ISSUE #452: Ensure front cam model class ordering is consistent with object_detection_utils

Rain1618:
## Changes Made
- Changed the class_names to be consistent with the classes of the YOLO model


MrMondrian (dismissed):

MrMondrian:
@Rain1618 can you also delete our api key from the ipynb?

MrMondrian:
@Rain1618 would you also be able to generate a new api key on roboflow if possible lol. just for security reasons because the old key is still in the git history

Rain1618:
> @Rain1618 would you also be able to generate a new api key on roboflow if possible lol. just for security reasons because the old key is still in the git history

Done!

antoinedang (approved):

PR #467, opened 2024-01-13
Vnc ubuntu

majdkhalife:
Added VNC functionality.


MrMondrian:
@antoinedang I worked with majd on this so I can vouch it is correct

MrMondrian (approved):

Issue #257, opened 2023-09-14
Change "confidence" in vision detections to be an attribute of the VisionObject message instead of a python array (makes for cleaner code)


antoinedang:

antoinedang:
nelsin started this, you can just finish (PR is open)

PR #479, opened 2024-01-18
made confidence an attribute of VisionObject class

MYRADhub:
## Related Issue(s)
  - Change "confidence" in vision detections to be an attribute of the VisionObject message instead of a python array (makes for cleaner code) #257

## Other Dependent PRs
 None

## Changes Made
- made confidence an attribute of VisionObject class
- changed code in object_detection.py and object_detection_utils.py to accommodate for the change

- to make the code more readable, I removed the confidence array and just used the existing DetectionFrameArray
- DetectionFrame.confidence is now the variable representing confidence of an object

## Tests Done
- Launched sim. No errors caused

## Optional Screen Recording
None

- new field in VisionObject.msg but I don't think it requires a wiki change

- None

- confidence-in-visionobject

### DIFF
--- a/catkin_ws/src/vision/src/object_detection.py
+++ b/catkin_ws/src/vision/src/object_detection.py
@@ -71,7 +71,7 @@ def detect_on_image(raw_img, camera_id):
             if camera_id == 0: # DOWN CAM
                 if global_class_name == "Lane Marker":
                     detectionFrame.label = global_class_name
-                    confidence.append(conf)
+                    detectionFrame.confidence = conf
                     headings, center, debug_img = measureLaneMarker(img, bbox, debug_img)
                     if None in headings:
                         detectionFrame.extra_field = None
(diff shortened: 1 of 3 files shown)


MrMondrian:
@MYRADhub when running sim.launch what did you do to ensure that your code ran and did not break? Also shouldn't you be working off the vision enhance branch?

MrMondrian (dismissed):

catkin_ws/src/vision/src/object_detection.py
```diff
@@ -71,7 +71,7 @@ def detect_on_image(raw_img, camera_id):
             if camera_id == 0: # DOWN CAM
                 if global_class_name == "Lane Marker":
                     detectionFrame.label = global_class_name
-                    confidence.append(conf)
+                    detectionFrame.confidence = conf
```
antoinedang, line 67: I think you forgot to delete the initialization of the confidence array (line 46)

PR #484, opened 2024-01-24
Hydrophone

frompotenza:
There is a PR in auv-sim-unity that depends on this


MrMondrian:
Its strange that the integration tests are failing here. @frompotenza @Rain1618 can one of you run them locally please

frompotenza:
> Its strange that the integration tests are failing here. @frompotenza @Rain1618 can one of you run them locally please

doesn't it fail all the time?

MrMondrian:
I fixed it a while ago

catkin_ws/src/auv_msgs/msg/PingerBearing.msg
antoinedang, line 1: It might make sense to include the AUV's pose in this message, or a timestamp maybe? What do you think @MrMondrian . It's probably not a big problem but it could help with ensuring consistency

MrMondrian (approved):

PR #519, opened 2024-02-10
Moved Camera-Related Launch Files to Sensors Package

Alb33rt:
## Related Issue(s)
  - Issue #380 

## Changes Made
- Moved stream-down-cam.launch & stream-front-cam.launch into package "sensors"
- Modified dependent parameters for these two files

- It makes sense for the camera "sensors" to be located in sensors

## Tests Done
- Displays the right thing on the sim?


catkin_ws/src/sensors/launch/stream-front-cam.launch
antoinedang, line 1: for these, the /vision/ relates to the topic name. We could change that to but it would require a change in a lot more places. @MrMondrian should we update the topic names u think

antoinedang: for the time being unless @MrMondrian corrects me I'd keep the topic names the same (so should be /vision/down_cam/... or whatever) @Alb33rt

MrMondrian: well we could just change the topic names everywhere

antoinedang: yeah the thing is that would mean updating topics in the sim, rviz and/or image view, a few other packages, as well as things like the camera info topics, point cloud, etc. I'd be down but its out of scope af so I think that makes more sense as a seperate issue

MrMondrian: lets put it back to vision then for now

Alb33rt: Ill just keep the topic names here the same and remove the changes then

Alb33rt:
So I reverted the topic name changes. There shouldn't be a problem with changing them everywhere else anymore.

PR #543, opened 2024-03-01
cleaned up mission substates

Malak-Oualid:
@antoinedang 
## Changes Made
- Added comments and renamed variables for more clarity


antoinedang (changes requested):
Hey Malak, looks good! I just would change one thing: in our codebase, the current convention is snake_case for variables and camelCase for functions. I see you changed everything to snake case, which is alright, but could we keep the functions in camelCase? Also, I'd make sure to test your code locally before making the PR (some of the functions which name you changed are defined in other files, so these missions would surely not work anymore)

Malak-Oualid:
Ok I'll update everything

antoinedang:
closed due to inactivity

PR #544, opened 2024-03-02
Object positions

Rain1618:
https://github.com/mcgill-robotics/AUV-2024/assets/115603054/e934b55a-b320-498a-aadd-86085a4d2b4a

## Changes Made
- Wrote a testing mission to check if the down cam can detect the positions of the lane marker and octagon table correctly.
- It takes like 10min for the test to run cause the AUV has trouble staying "settled". Attached sc below. Occasionally, everything works great and its done in like 30s tho. I'm not really sure what the problem could be tbh

image

- Could do a BFS instead of going to the known locations of each object directly.

## Optional Screen Recording
Attached below is a short clip of what happens when the auv is not settled and just kinda... drifts around. Also sc of the tests succeeding.
image


antoinedang (approved):
Looks good, great job

PR #560, opened 2024-04-02
new front cam

majdkhalife:


MrMondrian:
@majdkhalife have you had a chance to test the new model in sim?

antoinedang:
merging so we can test the new model, vision is pretty unusable right now

PR #569, opened 2024-05-01
Automation of image collection for training model (down cam and front…

frompotenza:
- simple script to facilitate collecting images (front and down cam at the same time) for when we decide to retrain vision models
- works for both sim and real environment


catkin_ws/src/vision/src/image_collection.py
```diff
@@ -0,0 +1,52 @@
...
+     if not os.path.exists(down_cam_data_dir):
+          os.makedirs(down_cam_data_dir)
+
+     front_cam_cur_image, down_cam_cur_image = None, None
+
+     front_cam_image_sub = rospy.Subscriber('/vision/front_cam/color/image_raw', Image, front_cam_image_callback) 
+     down_cam_image_sub = rospy.Subscriber('/vision/down_cam/image_raw', Image, down_cam_image_callback)
+     
+     print("These are the following options:\n- To take a screen shot using front cam, press [z]\n- To take a screen shot using down cam, press [x]")
+     while not rospy.is_shutdown():
```
MrMondrian, line 48: idk if this is gonna be a problem but you might need a better way to end the program than control c, could make the image not save properly. i know this is for sure a problem for text files

MrMondrian (approved):

catkin_ws/src/vision/src/image_collection.py
frompotenza: I used this script with docker, and I had no problems. Is this a problem for linux?

MrMondrian: it should be fine then

frompotenza:
@antoinedang can we merge this?

antoinedang (approved):
It'd be cool if later on you could make it so instead of typing x and then pressing enter we had a similar setup as the joystick and could just press X anytime and it would instantly take a screenshot (not at all important, nor a priority, just if you figure out how to get the joystick working in Docker then it would be useful to apply that here I think)

antoinedang:
this is the one i wanted to merge

frompotenza:
> It'd be cool if later on you could make it so instead of typing x and then pressing enter we had a similar setup as the joystick and could just press X anytime and it would instantly take a screenshot (not at all important, nor a priority, just if you figure out how to get the joystick working in Docker then it would be useful to apply that here I think)

I'll keep this in mind until I'm done with the keyboard

PR #574, opened 2024-05-29
Display improvements

frompotenza:
I tested this PR by moving the AUV around in the sim. There are a few things I want to improve but I think the core is working.

Hydrophones, Actuator, and front cam IMU have not been fully implemented/tested because the electrical team hasn't finished deciding the format of the data topic and/or we have not fully implemented it in the sim

There is a PR in the sim repo that depends on this, only merge this if the other PR is also good. 

Please don't ask me to write all of this in C++

There are a lot of changes in this PR, so please be very rigorous with what I did so that these changes don't become annoying in the future.

For some reason, there are some files where I only changed a few spaces and lines. I don't know how it happened since it looks unchanged on my VScode, but whatever


catkin_ws/src/sensors/package.xml
```diff
@@ -40,5 +42,7 @@
   <exec_depend>message_runtime</exec_depend>
   <exec_depend>python-bitstring-pip</exec_depend>
   <exec_depend>python3-serial</exec_depend>
+  <exec_depend>sensor_msgs</exec_depend>
+  <exec_depend>python3-numpy</exec_depend>
```
antoinedang, line 46: I've never seen us have to put python3-numpy, did you do this because it gave u an error or smth? Or can we remove

catkin_ws/src/state_estimation/src/state_aggregator.py
antoinedang, line 1: im not sure about combining the changes to state estimation with display board changes, makes the PR really hard to read. If its not too much work could you make 2 PRs (one for display board, one for state estimation changes) cuz I'm not sure where they're linked or where its just improvements/changing variable names

catkin_ws/src/sensors/package.xml
frompotenza: I kept getting an error related to the initialization of a np.quaternion object. So, when I looked at other packages that also use np.quaternion (sim, controls, state estimation, ...), they all had python3-numpy in their package.xml. I'm not 100% sure if this was the solution considering that I changed a bunch of things, but the error went away.

catkin_ws/src/state_estimation/src/state_aggregator.py
frompotenza: it's just that they are connected, but I can try to do a general overview:

- everything other than sensors.py and state_aggregator.py is for status only (changing message types to int32 instead of bool, adding launch files to run the status check, adding dependencies, and removing the status pub from the sim bridge)
- for all cpp files: I changed the message type from bool to int32, but refactoring everything in cpp was going to be a task too big for one person
- for sensors.py and state_aggregator.py: I can add comments in the actual code

catkin_ws/src/state_estimation/launch/state_estimation.launch
```diff
@@ -23,21 +23,21 @@
...
 		<param name="update_state_on_clock" value="false" />
 		<node name="state_aggregator" pkg="state_estimation" type="state_aggregator.py" respawn="true" output="screen"/>
-    </group>
-    <group if="$(arg sim)">
+	</group>
+	<group if="$(arg sim)">
 		<param name="update_state_on_clock" value="true" />
-		<node name="state_aggregator" pkg="state_estimation" type="state_aggregator" respawn="true" output="screen"/>
-    </group>
+		<node name="state_aggregator" pkg="state_estimation" type="state_aggregator.py" respawn="true" output="screen"/>
```
frompotenza, line 30: because I only refactored the python file, it would have to be .py

catkin_ws/src/state_estimation/src/sensors.py
```diff
@@ -19,111 +18,131 @@
```
frompotenza, line 19: Sensor() has a few getters() (we could also just retrieve the variable instead, but I think getters make it cleaner)

catkin_ws/src/state_estimation/src/sensors.py
```diff
@@ -19,111 +18,131 @@
...
         self.sensor_name = sensor_name
 
-        self.x = 0
-        self.y = 0
-        self.z = 0
-        
-        self.q_nwu_auv = np.quaternion(1, 0, 0, 0)
-        self.angular_velocity = np.array([0,0,0])
+    def get_sensor_name(self):
+        return self.sensor_name
```
frompotenza, line 26: get_sensor_name() just returns the name of the sensor so that It can be used when printing the warning messages in state aggregator

catkin_ws/src/state_estimation/src/sensors.py
```diff
@@ -19,111 +18,121 @@
...
 
-        # initialize a sensor as "inactive"
-        self.last_unique_state_time = -1 * self.time_before_considered_inactive
-        self.last_state = [self.x,self.y,self.z,self.q_nwu_auv,self.angular_velocity]
+    def sensor_status_cb(self, msg):
+        self.is_active = msg.data
+
```
frompotenza, line 30: sensor_status_cb gets the message from the corresponding sensor's status topic and saves it to is_active var

catkin_ws/src/state_estimation/src/sensors.py
```diff
@@ -19,111 +18,121 @@
...
+    # @abstractmethod
+    def get_reading(self):
+        raise NotImplementedError("Subclass must implement abstract method")
```
frompotenza, line 33: get_reading() is not related to status. It's a way to shorten the code so that we can avoid code duplication as it was before in state_aggregator. When implemented, it returns a dictionary with the format:
{
      "axis_name": axis_data
      ....(if there are more axis, add more entries)
} 
it also reduces instances of hard-code values in the implementation of state_aggregator

catkin_ws/src/state_estimation/src/sensors.py
```diff
@@ -19,111 +18,121 @@
...
-        if not self.hasValidData(): return False #check that state is complete
-        if rospy.get_time() - self.last_unique_state_time > self.time_before_considered_inactive: #check that state has changed in last N seconds
-            if rospy.get_time() - self.last_error_message_time > 1:
-                self.last_error_message_time = rospy.get_time()
-                rospy.logwarn("{} has been inactive for {} seconds.".format(self.sensor_name, self.time_before_considered_inactive))
-            return False
-        else:
-            return True
+    def get_is_active(self):
+        return self.is_active
```
frompotenza, line 36: just a getter to return whether the sensor is active or not. Used in state_aggregator

catkin_ws/src/state_estimation/src/sensors.py
```diff
@@ -19,111 +18,121 @@
...
-    def isActive(self):
-        if rospy.get_time() == 0: return False
-        if not self.hasValidData(): return False #check that state is complete
-        if rospy.get_time() - self.last_unique_state_time > self.time_before_considered_inactive: #check that state has changed in last N seconds
-            if rospy.get_time() - self.last_error_message_time > 1:
-                self.last_error_message_time = rospy.get_time()
-                rospy.logwarn("{} has been inactive for {} seconds.".format(self.sensor_name, self.time_before_considered_inactive))
-            return False
-        else:
-            return True
```
frompotenza, line 62: the warning messages were all moved to the sensor status, and I also reduced how frequent they were because they are annoying as fuck in the sim. When you are not moving the auv in the sim, the sensors readings are exactly the same. So, the status check thinks that they stopped working and starts spamming warning messages

catkin_ws/src/state_estimation/src/sensors.py
```diff
@@ -19,111 +18,121 @@
...
+    def get_is_active(self):
+        return self.is_active
 
-    def hasValidData(self):
-        raise NotImplementedError
```
frompotenza, line 65: moved to status check

catkin_ws/src/state_estimation/src/sensors.py
```diff
@@ -19,111 +18,121 @@
...
+    def __init__(self, sensor_name, is_active=0):
+        self.is_active = is_active
         self.sensor_name = sensor_name
 
-        self.x = 0
-        self.y = 0
-        self.z = 0
-        
-        self.q_nwu_auv = np.quaternion(1, 0, 0, 0)
-        self.angular_velocity = np.array([0,0,0])
```
frompotenza, line 32: I wasn't a big fan of having values for all axis even when the sensor doesn't provide data for that axis. So, I moved them to the subclass and, combined with get_reading(), I think it is a lot cleaner

catkin_ws/src/state_estimation/src/sensors.py
```diff
@@ -19,111 +18,121 @@
...
         q_imunominal_imu_z = rospy.get_param("q_imunominal_imu_z")
 
         self.q_imunominal_imu = np.quaternion(q_imunominal_imu_w, q_imunominal_imu_x, q_imunominal_imu_y, q_imunominal_imu_z)
         self.q_imu_auv = self.q_imunominal_imu.conjugate() * Q_IMUNOMINAL_AUV
-        self.q_nwu_auv = None
-        
+        self.q_nwu_auv = np.quaternion(1, 0, 0, 0)
+
+        self.angular_velocity = np.array([0,0,0])
+
```
frompotenza, line 74: these are just the initial values because None was causing problems, and I didn't want to have to add a bunch of if not None statements throughout the code

catkin_ws/src/state_estimation/src/sensors.py
```diff
@@ -19,111 +18,121 @@
...
         rospy.Subscriber("/sensors/imu/quaternion", SbgEkfQuat, self.quat_cb)
-        
+
     def ang_vel_cb(self, msg):
         # angular velocity vector relative to imu frame 
         ang_vel_imu = np.array([msg.gyro.x, msg.gyro.y, msg.gyro.z])
         # anuglar velocity vector relative to AUV frame 
         temp_angular_velocity = self.q_imu_auv * np.quaternion(0, ang_vel_imu[0], ang_vel_imu[1], ang_vel_imu[2]) * self.q_imu_auv.conjugate()
         self.angular_velocity = np.array([temp_angular_velocity.x, temp_angular_velocity.y, temp_angular_velocity.z])
-        self.updateLastState()
```
frompotenza, line 102: no need to update last reading given that the callback functions already do that

MrMondrian:
the only issue here is that state estimation is being refactored prior to pool testing where we may discover the entire package needs to be scrapped and rewritten anyway. its good work but might be wasted effort

catkin_ws/src/state_estimation/src/sensors.py
```diff
@@ -19,111 +18,121 @@
...
-        return self.q_nwu_auv is not None and self.angular_velocity is not None
+    def get_reading(self):
+        return {
+            "quaternion": self.q_nwu_auv, 
+            "angular_velocity": self.angular_velocity
+        }
 
+
+# IMUFrontCamera class inheriting from Sensor
+class IMUFrontCamera(Sensor):
```
frompotenza, line 99: add new imu front cam. The topic name for the data is supposed to work according to zed's website, but it's something I would have to confirm by using the actual camera

MrMondrian:
@frompotenza how did you test the changes to state estimation?

catkin_ws/src/state_estimation/src/sensors.py
```diff
@@ -19,111 +18,121 @@
...
+        self.q_nwu_auv = np.quaternion(1, 0, 0, 0)
+        self.angular_velocity = np.array([0,0,0])
+
+        rospy.Subscriber("/sensors/imu_front_camera/status", Int32, self.sensor_status_cb)
+        rospy.Subscriber("/zed/zed_node/imu/data", Imu, self.front_camera_imu_cb)
+
+    def front_camera_imu_cb(self, msg):
+        self.q_nwu_auv = np.quaternion(msg.orientation.w, msg.orientation.x, msg.orientation.y, msg.orientation.z)
+        self.angular_velocity = np.array([msg.angular_velocity.x, msg.angular_velocity.y, msg.angular_velocity.z]) 
+
```
frompotenza, line 112: I have no idea what the transformations are supposed to fix the reference frame, so I just set the variable to the new reading. We also need to check this in person

catkin_ws/src/state_estimation/src/sensors.py
```diff
@@ -19,111 +18,121 @@
...
+        return {
+            "quaternion": self.q_nwu_auv, 
+            "angular_velocity": self.angular_velocity
+        }
+
+
+# DVL class inheriting from Sensor
 class DVL(Sensor):
-    def __init__(self, imu):
+    def __init__(self, imu, imu_front_camera):
```
frompotenza: now that we have 2 imu's, we can use the backup as well

catkin_ws/src/state_estimation/src/sensors.py
```diff
@@ -132,60 +141,51 @@ def __init__(self, imu):
...
 
         if self.q_dvlref_nwu is None: return
 
         temp_pos_auv = self.q_dvlref_nwu.conjugate() * np.quaternion(0, pos_dvlref_dvl[0], pos_dvlref_dvl[1],pos_dvlref_dvl[2]) * self.q_dvlref_nwu
         pos_auv = np.array([temp_pos_auv.x, temp_pos_auv.y, temp_pos_auv.z])
-        temp_dvl_auv_offset_rotated = self.imu.q_nwu_auv * np.quaternion(0,self.auv_dvl_offset[0],self.auv_dvl_offset[1], self.auv_dvl_offset[2]) * self.imu.q_nwu_auv.conjugate()
+        if is_imu_active or not is_imu_front_camera_active:
+            temp_dvl_auv_offset_rotated = imu_q_nwu_auv * np.quaternion(0,self.auv_dvl_offset[0],self.auv_dvl_offset[1], self.auv_dvl_offset[2]) * imu_q_nwu_auv.conjugate()
+        else:
+            temp_dvl_auv_offset_rotated = imu_front_camera_q_nwu_auv * np.quaternion(0,self.auv_dvl_offset[0],self.auv_dvl_offset[1], self.auv_dvl_offset[2]) * imu_front_camera_q_nwu_auv.conjugate()
```
frompotenza: just adding some if's to decide if and which IMU to use

catkin_ws/src/state_estimation/src/sensors.py
```diff
@@ -132,60 +141,51 @@ def __init__(self, imu):
...
-        #     wx = rate_of_change_euler_y * math.sin(roll) * math.sin(yaw) + rate_of_change_euler_x * math.cos(yaw)
-        #     wy = rate_of_change_euler_y * math.sin(roll) * math.cos(yaw) - rate_of_change_euler_x * math.sin(yaw)
-        #     wz = rate_of_change_euler_y * math.cos(roll) + rate_of_change_euler_z
-
-        #     self.angular_velocity = np.array([wx, wy, wz])
-
-        #     self.roll = roll
-        #     self.pitch = pitch
-        #     self.yaw = yaw
-
```
frompotenza, line 187: fuck euler

MrMondrian: thank you i feel seen

catkin_ws/src/state_estimation/src/sensors.py
MrMondrian: tbh i would get rid of this. if the imu turns off id rather the auv just stop then do some unpredicatable behaviour

catkin_ws/src/state_estimation/src/state_aggregator.py
```diff
@@ -7,65 +7,75 @@
...
 
 DEG_PER_RAD = 180 / np.pi
 
+def swap_main_sensor_message(axis_index, sensor_index, sensor_name):
+    global last_warning_message_time
+
+    if sensor_index != 0 and rospy.get_time() - last_warning_message_time[axis_index] > sensor_swap_warning_interval:
+            last_warning_message_time[axis_index] = rospy.get_time()
+            rospy.logwarn("Using {} for {}.".format(sensor_name, axis_names[axis_index]))
+
```
frompotenza, line 26: this function is responsible for printing the warning message when the "main" (most important) sensor for a specific axis stops working and switches to another - not related to status check.

To avoid spamming, only print warning if the current sensor is not the main sensor (index_sensor is the index of the sensor in the sensor_priorities dictionary below, so index = 0 is the main sensor) AND it has been at least 5 seconds since the last warning 

the if index_sensor != 0 is outside the function so that this function doesn't have to be called every time

catkin_ws/src/state_estimation/src/state_aggregator.py
```diff
@@ -7,65 +7,75 @@
...
     if update_state_on_clock:
         global last_clock_msg_s
         global last_clock_msg_ns
         if last_clock_msg_s == msg.clock.secs and last_clock_msg_ns == msg.clock.nsecs: return
         last_clock_msg_s = msg.clock.secs
         last_clock_msg_ns = msg.clock.nsecs
-        
-    pub_dvl_sensor_status.publish(dvl.isActive())
-    pub_imu_sensor_status.publish(imu.isActive())
-    pub_depth_sensor_status.publish(depth_sensor.isActive())
```
frompotenza, line 26: status pub moved to sensors

catkin_ws/src/state_estimation/src/state_aggregator.py
```diff
@@ -7,65 +7,75 @@
...
-    x = None
-    y = None
-    z = None
+
+    position = [None, None, None]
```
frompotenza, line 36: again, avoid repetition of code - use indexes for position instead of different variables

catkin_ws/src/state_estimation/src/state_aggregator.py
```diff
@@ -7,65 +7,75 @@
...
-            break
-    for sensor in sensor_priorities["orientation"]:
-        if sensor.isActive():
-            quaternion = Quaternion(x = sensor.q_nwu_auv.x, y = sensor.q_nwu_auv.y, z = sensor.q_nwu_auv.z, w = sensor.q_nwu_auv.w)
-            euler_dvlref_dvl = transformations.euler_from_quaternion([sensor.q_nwu_auv.x, sensor.q_nwu_auv.y, sensor.q_nwu_auv.z, sensor.q_nwu_auv.w])
+    dvl_reading = dvl.get_reading()
+    imu_reading = imu.get_reading()
+    imu_front_camera_reading = imu_front_camera.get_reading()
+    depth_sensor_reading = depth_sensor.get_reading()
+
```
frompotenza: retrieve readings from all data at once

catkin_ws/src/state_estimation/src/state_aggregator.py
```diff
@@ -7,65 +7,75 @@
...
+    for axis_index in range(len(position)):
+        for sensor_index, sensor in enumerate(sensor_priorities[axis_names[axis_index]]):
+            if sensor.get_is_active():
+                position[axis_index] = sensor.get_reading()[axis_names[axis_index]]
+                swap_main_sensor_message(axis_index, sensor_index, sensor.get_sensor_name()) 
+                break
+
```
frompotenza, line 42: same idea as before but now we have less repetition
outer loop iterates over x, y, z axis 
inner loop iterates over each sensor (from sensor_priorities) in the corresponding axis 
if the sensor is active - use reading. If it is not the main one that is active, go to warning message

catkin_ws/src/state_estimation/src/state_aggregator.py
```diff
@@ -7,65 +7,76 @@
...
+    for sensor_index, sensor in enumerate(sensor_priorities["orientation"]):
+        if sensor.get_is_active():
+            sensor_reading = sensor.get_reading()
+            reading_quaternion = sensor_reading["quaternion"]
+            reading_angular_velocity = sensor_reading["angular_velocity"] 
+            
+            quaternion = Quaternion(x = reading_quaternion.x, y = reading_quaternion.y, z = reading_quaternion.z, w = reading_quaternion.w)
+            euler_dvlref_dvl = transformations.euler_from_quaternion([reading_quaternion.x, reading_quaternion.y, reading_quaternion.z, reading_quaternion.w])
```
frompotenza, line 54: I didn't add orientation to the previous for loop because it uses 2 data readings instead of just one, so I think it's just better to leave it as it is

frompotenza:
> the only issue here is that state estimation is being refactored prior to pool testing where we may discover the entire package needs to be scrapped and rewritten anyway. its good work but might be wasted effort

wait what? why would it need to be scrapped?

catkin_ws/src/state_estimation/src/sensors.py
frompotenza: what if the front camera IMU is somewhat good?

frompotenza:
> @frompotenza how did you test the changes to state estimation?

since I only changed sensors stuff, I moved the auv around in the sim to see how the display would change and the messages being printed from state_aggregator. It was working as expected

catkin_ws/src/sensors/package.xml
antoinedang: ive had a similar problem in the past that was related to the order of importing numpy/quaternion (first numpy then quaternion)

catkin_ws/src/state_estimation/src/sensors.py
```diff
@@ -19,111 +18,121 @@
...
-    def hasValidData(self):
-        raise NotImplementedError    
 
+# Depth Sensor class inheriting from Sensor
 class DepthSensor(Sensor):
     def __init__(self):
         super().__init__("Depth Sensor")
+        
+        self.z = 0
         self.z_pos_mount_offset = 0
```
antoinedang: can we make z_pos_mount_offset a ros param?

catkin_ws/src/state_estimation/src/sensors.py
antoinedang: I don't like the ifs either but for a different reason, since we already have a tree of sensors and their priorities in state aggregator I would pass that into DVL (pass a callable into DVL or just setter for the current IMU or something)

antoinedang: its fair that if IMU dies DVL should die. Regardless tho overall DVL class is gonna change a ton so i dont really care tbh

catkin_ws/src/state_estimation/src/state_aggregator.py
```diff
@@ -7,65 +7,72 @@
...
 from geometry_msgs.msg import Pose, Quaternion, Vector3, TransformStamped, Point
-from sensors import DepthSensor, IMU, DVL
-from std_msgs.msg import Float64, Bool
+from sensors import DepthSensor, IMU, IMUFrontCamera, DVL
+from std_msgs.msg import Float64, Int32
 from rosgraph_msgs.msg import Clock
 
 DEG_PER_RAD = 180 / np.pi
 
+def swap_main_sensor_message(axis_index, sensor_name):
```
antoinedang: the function name had me hella confused, could we just call it "print_sensor_source" or something? Also im not sure if its best to have it check for 5 seconds inside the function but check for axis !=0 outside, might as well have all logic related to logging inside (function calls take like 0.0000001 seconds)

antoinedang (approved):

antoinedang:
merge manually cuz there are 50 unresolved convos blocking merge

PR #576, opened 2024-05-29
Docker reorganizing

MrMondrian:
## Related Issue(s)
  - Issue #1
  - Issue #2

## Other Dependent PRs
  - PR #2
  - PR #3

## Changes Made
- Describe the changes made in this PR.

- Explain the rationale behind this PR.
- Mention any alternative solutions considered (if any).

## Tests Done
- Describe the tests that have been performed.
- Mention any relevant test cases or scenarios.

## Optional Screen Recording
Attach a screen recording of test run(s) (if applicable).

- List any changes that need to be made to the project's wiki (if applicable).

- List any new dependencies that need to be installed by the install_dependencies.sh script

- List any related branches (if applicable).


frompotenza:
this might be a record for number of files changed

MrMondrian:
> this might be a record for number of files changed

yeah probably

catkin_ws/src/state_estimation/launch/state_estimation.launch
```diff
@@ -26,7 +26,7 @@
     </group>
     <group if="$(arg sim)">
 		<param name="update_state_on_clock" value="true" />
-		<node name="state_aggregator" pkg="state_estimation" type="state_aggregator" respawn="true" output="screen"/>
+		<node name="state_aggregator" pkg="state_estimation" type="state_estimation" respawn="true" output="screen"/>
```
antoinedang: how does this work? (genuine question) I thought that .cpp files in state estimation would be compiled as executables and then we run those, so how is state_estimation an executable which is built?

catkin_ws/src/sensors/package.xml
```diff
@@ -25,7 +25,6 @@
   <build_depend>message_generation</build_depend>
   <build_depend>rospy</build_depend>
   <build_depend>sbg_driver</build_depend>
-  <build_depend>zed_wrapper</build_depend>
```
antoinedang, line 28: the sensors package does build_depend on zed_wrapper though I think

Docker/poolside/build_zed.bash
antoinedang, line 1: smart af

Docker/base/Dockerfile
```diff
@@ -1,6 +1,14 @@
 # Use the official ROS Noetic base image
 FROM ros:noetic-ros-base
 
+RUN apt-get update && apt-get install -y locales\
```
antoinedang, line 4: only change I would make is put ultralytics + scikit in base image (if it doesnt cause problems with the actions) cuz it is more of a base feature than a sim-specific thing. Would also remove the need for a sim image and we could just have base or main or whatever, poolside/jetson, and actions. And also it enables us to have vision integration testing later on if we want

catkin_ws/src/state_estimation/launch/state_estimation.launch
MrMondrian: in the cmakelists add_executable function i just set it to the project name variable instead of naming it something special

Docker/base/Dockerfile
MrMondrian: thats valid but notice that the actions aren't slow anymore

catkin_ws/src/state_estimation/launch/state_estimation.launch
MrMondrian: think of how you compile with gcc. 
```bash
gcc a.cpp -o b
```
the resulting executable can be anything. its the same in cmake. instead of requesting it be called state_aggregator i just passed it the value of the project variable which is fairly standard i believe

MrMondrian:
@antoinedang the poolside image gives this error
`CUDA error at Camera.cpp:293 code=804(cudaErrorCompatNotSupportedOnDevice) "void sl::Camera::close()" ` when running the ros wrapper. no idea if this is actually a cuda issue or its because im running a camera driver without a camera. seeing as that the error in being thrown in a camera related file in a camera related method good chance this will work when the camera is connected

catkin_ws/src/state_estimation/launch/state_estimation.launch
antoinedang: ah ok i didnt realize u had set that in cmake gotcha thanks

antoinedang:
> @antoinedang the poolside image gives this error `CUDA error at Camera.cpp:293 code=804(cudaErrorCompatNotSupportedOnDevice) "void sl::Camera::close()" ` when running the ros wrapper. no idea if this is actually a cuda issue or its because im running a camera driver without a camera. seeing as that the error in being thrown in a camera related file in a camera related method good chance this will work when the camera is connected

ok lets hold off on merging then and we can test this on the real jetson this weekend or something

Docker/base/Dockerfile
antoinedang: was it that slow because of ultralytics? can we try pushing with ultralytics in base image and seeing how long it takes? cuz if its not a significant slow down then makes sense to have it in base image imo

PR #582, opened 2024-06-15
Hydrophones new type integration

frompotenza:
There is also a PR in the sim repo

- Once electrical figures out all the details (e.g., real frequency values, possible overflow problem), I can open another PR to change it. But for now, there are no errors

### DIFF
--- a/catkin_ws/src/auv_msgs/msg/PingerBearing.msg
+++ b/catkin_ws/src/auv_msgs/msg/PingerBearing.msg
@@ -1,4 +1,4 @@
-int32 frequency_index
+int32 frequency
 geometry_msgs/Vector3 pinger_bearing
 
 float64 state_x
--- a/catkin_ws/src/auv_msgs/msg/PingerBearing.msg
+++ b/catkin_ws/src/auv_msgs/msg/PingerBearing.msg
@@ -1,7 +1,5 @@
-geometry_msgs/Vector3 pinger1_bearing
-geometry_msgs/Vector3 pinger2_bearing
-geometry_msgs/Vector3 pinger3_bearing
-geometry_msgs/Vector3 pinger4_bearing
+int32 frequency_index
+geometry_msgs/Vector3 pinger_bearing
 
 float64 state_x
 float64 state_y
--- a/catkin_ws/src/sensors/launch/sensors_status.launch
+++ b/catkin_ws/src/sensors/launch/sensors_status.launch
@@ -1,5 +1,10 @@
 <launch>
      <param name="sensor_warning_interval" value="5"/> <!-- secs -->
      <param name="sensor_status_update_rate" value="10"/> <!-- Hz -->
+     <!-- Frequency of each pinger -->
+	<param name="pinger_frequency_1" value="1"/>
+	<param name="pinger_frequency_2" value="2"/>
+	<param name="pinger_frequency_3" value="3"/>
+	<param name="pinger_frequency_4" value="4"/>
      <node name="sensors_status" pkg="sensors" type="sensors_status.py" output="screen" respawn="true"/>
 </launch>
--- a/catkin_ws/src/sensors/launch/sensors_status.launch
+++ b/catkin_ws/src/sensors/launch/sensors_status.launch
@@ -1,6 +1,7 @@
 <launch>
      <param name="sensor_warning_interval" value="5"/> <!-- secs -->
      <param name="sensor_status_update_rate" value="10"/> <!-- Hz -->
+     <param name="time_before_considered_inactive" value="5"/>
      <!-- Frequency of each pinger -->
 	<param name="pinger_frequency_1" value="1"/>
 	<param name="pinger_frequency_2" value="2"/>
--- a/catkin_ws/src/sensors/launch/sensors_status.launch
+++ b/catkin_ws/src/sensors/launch/sensors_status.launch
@@ -2,10 +2,6 @@
      <param name="sensor_warning_interval" value="5"/> <!-- secs -->
      <param name="sensor_status_update_rate" value="10"/> <!-- Hz -->
      <param name="time_before_considered_inactive" value="5"/>
-     <!-- Frequency of each pinger -->
-	<param name="pinger_frequency_1" value="1"/>
-	<param name="pinger_frequency_2" value="2"/>
-	<param name="pinger_frequency_3" value="3"/>
-	<param name="pinger_frequency_4" value="4"/>
+
      <node name="sensors_status" pkg="sensors" type="sensors_status.py" output="screen" respawn="true"/>
 </launch>
--- a/catkin_ws/src/sensors/src/sensors_status.py
+++ b/catkin_ws/src/sensors/src/sensors_status.py
@@ -234,12 +234,15 @@ class Hydrophones(Sensor):
         # @TODO(Felipe): Subsbitute frequency_types values by 
         #                real pinger values (waiting for Sam
         #                to respond)  
-        self.frequency_types = [
-            rospy.get_param("pinger_frequency_1"), 
-            rospy.get_param("pinger_frequency_2"), 
-            rospy.get_param("pinger_frequency_3"), 
-            rospy.get_param("pinger_frequency_4")
-        ]
+        if rospy.has_param("pinger_frequency_1"):
+            self.frequency_types = [
+                rospy.get_param("pinger_frequency_1"), 
+                rospy.get_param("pinger_frequency_2"), 
+                rospy.get_param("pinger_frequency_3"), 
+                rospy.get_param("pinger_frequency_4")
+            ]
+        else:
+            self.frequency_types = [-1, -1, -1, -1]
         self.current_reading = [None] * len(self.frequency_types)
         self.last_reading = [None] * len(self.frequency_types)
         self.frequency_index = -1
--- a/catkin_ws/src/sensors/src/sensors_status.py
+++ b/catkin_ws/src/sensors/src/sensors_status.py
@@ -18,11 +18,11 @@ RAD_PER_DEG = 1 / DEG_PER_RAD
 
 class Sensor:
     def __init__(self, sensor_name):
-        self.time_before_considered_inactive = 1  # seconds
+        self.time_before_considered_inactive = rospy.get_param("time_before_considered_inactive")
         self.sensor_name = sensor_name
         self.sensor_warning_interval = rospy.get_param(
             "sensor_warning_interval"
-        )  # seconds
+        )
 
         # initialize a sensor as "inactive"
         self.last_unique_reading_time = -1 * self.time_before_considered_inactive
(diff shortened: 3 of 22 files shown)


catkin_ws/src/auv_msgs/msg/PingerBearing.msg
```diff
@@ -1,7 +1,5 @@
-geometry_msgs/Vector3 pinger1_bearing
-geometry_msgs/Vector3 pinger2_bearing
-geometry_msgs/Vector3 pinger3_bearing
-geometry_msgs/Vector3 pinger4_bearing
+int32 frequency_index
```
antoinedang: I think it would be clearer to keep it as an overall frequency and not an index, will prevent mistakes in the future if we decide to change the order or whatever and also itll make sure the message is more or less standalone

catkin_ws/src/sensors/launch/sensors_status.launch
```diff
@@ -1,5 +1,10 @@
 <launch>
      <param name="sensor_warning_interval" value="5"/> <!-- secs -->
      <param name="sensor_status_update_rate" value="10"/> <!-- Hz -->
+     <!-- Frequency of each pinger -->
+	<param name="pinger_frequency_1" value="1"/>
```
antoinedang: these will be like 20_000, 25_000 in the future right?

antoinedang: lets just put them to what they will be at comp right now tbh

catkin_ws/src/sensors/src/sensors_status.py
```diff
@@ -242,31 +251,20 @@ def __init__(self):
...
             self.last_unique_reading_time = rospy.get_time()
             self.last_reading = self.current_reading[:]
 
     def has_valid_data(self):
-        indexes_true = [i for i, e in enumerate(self.are_pingers_active) if e]
-        indexes_non_zero = [
-            i for i, e in enumerate(self.current_reading) if e != [0, 0, 0, 0]
-        ]
-        return len(indexes_true) != 0 and indexes_true == indexes_non_zero
+        return True
```
antoinedang: probably a leftover from debugging, we should uncomment before merge

catkin_ws/src/state_estimation/launch/hydrophones.launch
```diff
@@ -0,0 +1,20 @@
...
+	<param name="hydrophones_dx" value="0.1"/>
+	<param name="hydrophones_dy" value="0.1"/>
+	<param name="hydrophones_dz" value="-0.1"/>
+	<!-- Frequency of each pinger -->
+	<param name="pinger_frequency_1" value="1"/>
+	<param name="pinger_frequency_2" value="2"/>
+	<param name="pinger_frequency_3" value="3"/>
+	<param name="pinger_frequency_4" value="4"/>
+
+	<param name="hydrophones_time_unit" value="0.0000001"/>
```
antoinedang, line 12: this is kinda confusing, what does this represent?

catkin_ws/src/auv_msgs/msg/PingerBearing.msg
frompotenza: yeah but then we have to keep loading the same parameters multiple times, turning them into a list and then getting the index. This way, you can skip steps

catkin_ws/src/sensors/launch/sensors_status.launch
frompotenza: do you know what they are? I couldn't find the exact values in the competition document. They only give a range. I was hoping electrical would know

antoinedang: ah ok i thought we had the values, but yeah I checked last year's and they said the values changed every day and were announced at competition

catkin_ws/src/auv_msgs/msg/PingerBearing.msg
antoinedang: I'm confused at what level this helps us gain steps? When we receive the message, we'll need to load an array and index it to know which frequency the index represents. If we use indexes we'll need to maintain an array of frequencies across the repo that will have to change based on the order of tasks, if they change frequencies, etc. If we keep everything in terms of frequencies, we don't have to keep track of any ordering, we can just know which direction certain frequencies point, and then just have a ros param for each task that indicates its frequency.

frompotenza: Let's say you are writing the planer and the task needs a pinger. By using indexes, you don't have to worry about the actual value of the frequency. You can just say, the first task that uses a pinger is index 0, the second is index 1, and so on. If we only use frequencies, you would have to change the planner, state estimation, and sensors and add if-statements in the planner to get the correct frequency for a particular task every time the frequency values change

catkin_ws/src/sensors/src/sensors_status.py
frompotenza: I fix it with the latest commit

catkin_ws/src/state_estimation/launch/hydrophones.launch
frompotenza: which part?

antoinedang:
@frompotenza I'm putting this as a draft PR, ping me on discord when u want me to review it / merge

catkin_ws/src/state_estimation/src/hydrophones_bearing.py
```diff
@@ -1,269 +1,116 @@
...
-    dx = c * dtx
-    dy = c * dty
-    dz = c * dtz
-    return [dx, dy, dz]
+    hydrophones_dx = rospy.get_param("hydrophones_dx")
+    hydrophones_dy = rospy.get_param("hydrophones_dy")
+    hydrophones_dz = rospy.get_param("hydrophones_dz")
 
+    frequency_types = [
+            rospy.get_param("pinger_frequency_1"),
```
antoinedang, line 80: dont need this anymore right?

catkin_ws/src/state_estimation/src/hydrophones_bearing.py
```diff
@@ -1,269 +1,116 @@
...
-            )
-            bearing_vector_global = quaternion.rotate_vectors(
-                auv_rotation,
-                np.array([bearing_vector_local[0], bearing_vector_local[1], 0]),
-            )
-            PingerBearing_msg.pinger3_bearing.x = bearing_vector_global[0]
-            PingerBearing_msg.pinger3_bearing.y = bearing_vector_global[1]
-            PingerBearing_msg.pinger3_bearing.z = 0
+def cb_hydrophones_time_difference(msg):
+    if not is_active or msg.frequency not in frequency_types:
```
antoinedang: dont need to check if frequency is valid, should just publish all frequencies (they'll always be one of the ones we define as possible)

antoinedang: although could be a useful sanity check, but if its a sanity check it needs to be verbose and not silently fail

catkin_ws/src/state_estimation/launch/hydrophones.launch
```diff
@@ -0,0 +1,20 @@
+<launch>
+     <!-- Distances to hydrophone 1 -->
+	<param name="hydrophones_dx" value="0.1"/>
+	<param name="hydrophones_dy" value="0.1"/>
+	<param name="hydrophones_dz" value="-0.1"/>
+	<!-- Frequency of each pinger -->
+	<!-- <param name="pinger_frequency_1" value="1"/>
```
antoinedang: can just remove rather than comment out

antoinedang: although actually why is this commented out isnt it still used in the bearing calculation file

antoinedang: I think it makes more sense to be in hydrophones than sensor status

catkin_ws/src/sensors/launch/sensors_status.launch
```diff
@@ -1,5 +1,11 @@
 <launch>
      <param name="sensor_warning_interval" value="5"/> <!-- secs -->
      <param name="sensor_status_update_rate" value="10"/> <!-- Hz -->
+     <param name="time_before_considered_inactive" value="5"/>
+     <!-- Frequency of each pinger -->
+	<param name="pinger_frequency_1" value="1"/>
```
antoinedang: see comment below, should be in hydrophones.launch

catkin_ws/src/state_estimation/launch/hydrophones.launch
frompotenza: yeah, I just did. The params are loaded in sensor_status.launch. No need to load it twice

catkin_ws/src/state_estimation/src/hydrophones_bearing.py
frompotenza: It's just to make sure that if electrical fucks it up, we don't get fucked as well

catkin_ws/src/state_estimation/launch/hydrophones.launch
antoinedang: not sure if u saw the comment, it makes more sense to have pinger frequencies in hydrophones.launch no? If we want to launch hydrophones without sensor status / sensors currently it will crash. Neither should crash, but sensor status can be launched without hydrophones and that makes sense: so we can just add a try/catch on the parameter getting with maybe a while loop so that it crashes elegantly and with a clear error message

catkin_ws/src/state_estimation/src/hydrophones_bearing.py
frompotenza: sensor status checks if the frequency received is one of the correct frequencies, but I was worried in case there is some delay and the is_active isn't updated in time to stop the bearing  calculations

frompotenza: it doesn't have to be verbose because we will see the display turning red

antoinedang: yeah no i think ur right to check, just add a print/error log statement or something so it doesnt fail silently. but yeah i agree with you actually it makes sense to double check

antoinedang: github comments smh

antoinedang: we should still have a print statement though, there's 100 reasons the display could turn red and in general we shouldn't rely entirely on the display, its just useful as a quicker debugging tool

catkin_ws/src/state_estimation/launch/hydrophones.launch
frompotenza: but then we would have to launch state_estimation before sensor_status, because they both use the same frequency param. Or we could load the same params in two different places

frompotenza:
@antoinedang it's ready now

antoinedang (approved):

PR #592, opened 2024-06-29
Joystick pid

antoinedang:
THIS IS NOT A SERIOUS PR, PLEASE DON'T APPROVE IT


antoinedang:
just making this to see how the joystick is supposed to work

catkin_ws/src/controls/src/joystick.py
```diff
@@ -1,85 +1,472 @@
...
+            if self.wait_execute:
+                rospy.sleep(2)
+            current_time = rospy.get_time()
+            if (
+                any(x is None for x in [self.x, self.y, self.z, self.quaternion]) or 
+                current_time - self.last_pose_received > self.missing_pose_interval
+            ):
+                if current_time - self.last_warning_time > self.missing_pose_interval:
+                    self.last_warning_time = current_time 
+                    rospy.logwarn("Missing pose values.")
```
antoinedang, line 441: would be helpful to print which ones

frompotenza:
> just making this to see how the joystick is supposed to work

what?

catkin_ws/src/controls/src/joystick.py
frompotenza: not really because state estimation will just send 0s

antoinedang: whats the point of this check then? checking last pose received is all we need to do not check if x/y/z/quat are none

frompotenza: I mean, I haven't finished writing the code yet. I put this as a way to remember to add some pose-check and give a warning since I use PIDs. But this branch is still not ready to be a PR

PR #593, opened 2024-07-05
Vision Refactoring + Lane Marker tests

frompotenza:
Tested in the sim, and it still works as before


catkin_ws/src/vision/src/render_visualization.py
```diff
@@ -1,153 +1,152 @@
...
         "Surge:0\nSway:0\nHeave:0\nRoll:0\nPitch:0\nYaw:0",
-        auv_pub.publish,
+        pub_auv.publish,
         0.15,
-    )  # 10 - effort
-    addHeading(-1000, 0, 0, 1, 0, 0, auv_pub.publish, (1, 1, 1))  # 11 - setpoint x axis
-    addHeading(-1000, 0, 0, 0, 1, 0, auv_pub.publish, (1, 1, 1))  # 12 - setpoint y axis
-    addHeading(-1000, 0, 0, 0, 0, 1, auv_pub.publish, (1, 1, 1))  # 13 - setpoint z axis
+    )  # 10 - effort.
+    add_heading(-1000, 0, 0, 1, 0, 0, pub_auv.publish, (1, 1, 1))  # 11 - setpoint x axis.
```
antoinedang: we can just delete this file, visualization is done in unity now

antoinedang:
is it possible to split this up into smaller PRs? This is too massive to review properly (i.e. one PR for refactors, a seperate one for tests)

MrMondrian:
@frompotenza lgtm. I think a good think for readability in vision would be to make all params uppercase so its clear which values are variables or hyper parameters

frompotenza:
> is it possible to split this up into smaller PRs? This is too massive to review properly (i.e. one PR for refactors, a seperate one for tests)

almost nothing is for tests, only the stuff in the tests folder

frompotenza:
> @frompotenza lgtm. I think a good think for readability in vision would be to make all params uppercase so its clear which values are variables or hyper parameters

Do you mean rosparams?

antoinedang:
> > @frompotenza lgtm. I think a good think for readability in vision would be to make all params uppercase so its clear which values are variables or hyper parameters
> 
> Do you mean rosparams?

I'd assume so, anything thats not a variable. I agree with that idea too, it makes things easier to read

Other than that I'm good to merge if you've tested and its working as good as without these changes. Please 😢 🙏 I don't want to debug vision functions anymore

PR #604, opened 2024-08-07
Octagon Task

Rain1618:
## Changes Made
- Uses NavigatePinger to approach the octagon
- Then uses front and down cam to center the AUV over the octagon table and surface

- Make AUV do a lil in place search to determine the position of the octagon table

## Tests Done
- Works in (felipe's) sim (mine is still broken)


catkin_ws/src/vision/launch/vision.launch
```diff
@@ -65,7 +65,7 @@
     <param name="min_distance_for_valid_point_cloud_point" value="0.5" />
     <param name="max_counts_per_label" value='{"Buoy":1, "Gate":1, "Lane Marker":2, "Octagon Table":1, "Bin":1}'/>
     <param name="debug_lane_marker_thresholding" value="false" />
-    <param name="debug_point_cloud_cleaning" value="false" />
+    <param name="debug_point_cloud_cleaning" value="true" />
```
frompotenza: Why did you change this?

catkin_ws/src/planner/src/substates/linear_search.py
```diff
@@ -8,10 +8,11 @@
 
 # ASSUMES AUV IS FACING DIRECTION TO SEARCH IN
 class LinearSearch(smach.State):
-    def __init__(self, control, mapping, target_class, min_objects):
+    def __init__(self, control, mapping, state, target_class, min_objects):
```
frompotenza: why do you need state here? It doesn't seem to be used anywhere

catkin_ws/src/planner/src/missions.py
```diff
@@ -237,17 +239,30 @@ def octagon(
...
+                "success": target_success_state_name,
                 "timeout": target_timeout_state_name,
                 "failure": "failure",
             },
         )
+
         smach.StateMachine.add(
             second_state_name,
+            LinearSearch(
+                self.control, self.mapping, self.state, target_class="Octagon Table", min_objects=1
```
frompotenza: state change here as well

catkin_ws/src/planner/src/missions.py
```diff
@@ -237,17 +239,30 @@ def octagon(
 
         smach.StateMachine.add(
             first_state_name,
-            LinearSearch(
-                self.control, self.mapping, target_class="Octagon Table", min_objects=1
+            NavigatePinger(
+                self.control, self.mapping, self.state,
             ),
```
frompotenza, line 254: octagon shouldn't have navigate pinger. Just assume that you are close enough to the octagon that you can do a regular search

catkin_ws/src/planner/src/substates/octagon_task.py
```diff
@@ -33,28 +33,25 @@ def execute(self, ud):
...
             return "failure"
 
         if self.timeout_occurred:
             return "timeout"
+        
         print("Moving to the center of the octagon.")
         self.control.move(
             (octagon_obj[1], octagon_obj[2], rospy.get_param("down_cam_search_depth")),
             face_destination=True,
         )
```
frompotenza: It looks really good. The only problem is that it doesn't wait for the object map to improve. Rn, it sets the pid after the first detection. But the first detection does not always translate to a good object map. So, you can either wait a bit for the object map to update and correct itself or update the pid as the object map improves (with some loop)

frompotenza: After testing in the sim, it sometimes kills the thrusters too early and it ends up hitting the sides of the octagon instead of being in the middle.

frompotenza (changes requested):

catkin_ws/src/planner/src/substates/octagon_task.py
Rain1618: Made a loop that gets the location of the octagon as the AUV moves towards it then breaks when the AUV is close enough to the octagon that we think it can accurately detect its location. Currently that threshold is set to 3. Also fixed typos in object_map.py

PR #611, opened 2024-08-16
updated drytest to select whatever thrusters you want continuously

NathanielFactor:
## Related Issue(s)
  - Issue #1
  - Issue #2

## Other Dependent PRs
  - PR #2
  - PR #3

## Changes Made
- Describe the changes made in this PR.

- Explain the rationale behind this PR.
- Mention any alternative solutions considered (if any).

## Tests Done
- Describe the tests that have been performed.
- Mention any relevant test cases or scenarios.

## Optional Screen Recording
Attach a screen recording of test run(s) (if applicable).

- List any changes that need to be made to the project's wiki (if applicable).

- List any new dependencies that need to be installed by the install_dependencies.sh script

- List any related branches (if applicable).

### DIFF
--- a/catkin_ws/src/propulsion/src/drytest.py
+++ b/catkin_ws/src/propulsion/src/drytest.py
@@ -52,6 +52,14 @@ def simultaneous_forwards_test():
         if choice != "1":
             break
 
+# just type the thruster you want and it will run
+def optimized_dry_test(t):
+    print("- spinning at " + str(100 * force_amt) + "% max forwards force for 1s")
+    cmd = reset.copy()
+    cmd[t - 1] = force_to_pwm(force_amt * MAX_FWD_FORCE * thruster_mount_dirs[t-1])
+    pub.publish(cmd)
+    rospy.sleep(1.0)
+    pub.publish(reset_cmd)
 
 def reset_thrusters():
     pub.publish(reset_cmd)
@@ -97,7 +105,7 @@ while not rospy.is_shutdown():
         while True:
             choice = int(input("select thruster to test (1-8): "))
             if 1 <= choice and choice <= 8:
-                forwards_test(choice)
+                optimized_dry_test(choice)
             else:
                 break
     elif choice == "3":
--- a/catkin_ws/src/propulsion/src/drytest.py
+++ b/catkin_ws/src/propulsion/src/drytest.py
@@ -94,9 +94,12 @@ while not rospy.is_shutdown():
         forwards_test(7)
         forwards_test(8)
     elif choice == "2":
-        choice = int(input("select thruster to test (1-8): "))
-        if 1 <= choice and choice <= 8:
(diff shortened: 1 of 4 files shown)


catkin_ws/src/propulsion/src/drytest.py
```diff
@@ -94,9 +94,12 @@ def re_arm():
         forwards_test(7)
         forwards_test(8)
     elif choice == "2":
-        choice = int(input("select thruster to test (1-8): "))
-        if 1 <= choice and choice <= 8:
-            forwards_test(choice)
+        while True:
+            choice = int(input("select thruster to test (1-8): "))
+            if 1 <= choice and choice <= 8:
+                forwards_test(choice)
```
frompotenza: I think you can simplify it so that it doesn't use the forward_test function. Maybe just run whatever number is pressed (e.g., if you want to run thruster 7 twice and then thruster 8 twice, the input would be 7788 instead of 712812). This way, you could fully remove the first if statement (if choice == "1") in the main while loop - redundant code.

NathanielFactor: ok ill test tn w new version

NathanielFactor:
fixed CI w new repo name, also committed new changes to dry test. Should be good now I think, can't test in sim until later tonight but I think its fine.

catkin_ws/src/propulsion/src/drytest.py
```diff
@@ -94,9 +102,12 @@ def re_arm():
         forwards_test(7)
```
frompotenza: Can't you fully remove this first option (if) and the forwards_test function? If you want to do 1-8 in sequence, you can just press 1-2-3-4-5-6-7-8 on your keyboard. No need to keep pressing next for forward_test

NathanielFactor: Yes, but I think this is nice to have for communication between the people on the thrusters and us who are running the tests. I thought it may be nice that we have it so we can tell them "Running 1, running 2, repeating 2, running 5, etc." without having the restriction of going in order. If we get rid of the if, it will just run itself without us communicating "okay 7 is running, 5 is running". Sure, you can add some print statements but a lot of the time they will be like "wait run that back again", then you would have to wait until your sequence is done until you can.

NathanielFactor: Also, i am not using forwards test, I created a new function optimized_test() so maybe u r looking at wrong commit

NathanielFactor: Oh I see what you mean, just get rid of forwards test all together and the first dry test option. My bad, I will do that right now.

frompotenza: nw, i'm just suggesting changes. If you think there is a better way, feel free to do it

PR #691, opened 2025-05-04
New thruster mapping based on thruster test of May 2025

nadimasmar:
This PR introduces a complete refactor of the thruster_mapper and thruster_mapper_utils modules to integrate per-thruster force-to-PWM conversion functions, based on real-world test data collected in May 2025.

Changes:
Replaced the single force_to_pwm() function with dedicated polynomial-based mappings for each of the 8 thrusters (force_to_pwm_thruster1 to force_to_pwm_thruster8).

Each function uses a two-sided 5th order polynomial fit, customized per thruster using measured thrust vs. PWM data.

Added linear smoothing in the deadband region (±0.5 N) to reduce jitter near neutral.

Updated thruster_mapper.py to:

Call the appropriate function per thruster

### DIFF
--- a/catkin_ws/src/propulsion/src/thrust_mapper.py
+++ b/catkin_ws/src/propulsion/src/thrust_mapper.py
@@ -125,14 +125,14 @@ class ThrusterMapper:
         Applies individual limits to prevent overcurrent.
         """
         pwm_arr = [None] * 8
-        pwm_arr[ThrusterMicroseconds.BACK_LEFT] = force_to_pwm_thruster8(forces_msg.BACK_LEFT * thruster_mount_dirs[ThrusterMicroseconds.BACK_LEFT])
-        pwm_arr[ThrusterMicroseconds.HEAVE_BACK_LEFT] = force_to_pwm_thruster7(forces_msg.HEAVE_BACK_LEFT * thruster_mount_dirs[ThrusterMicroseconds.HEAVE_BACK_LEFT])
-        pwm_arr[ThrusterMicroseconds.HEAVE_FRONT_LEFT] = force_to_pwm_thruster6(forces_msg.HEAVE_FRONT_LEFT * thruster_mount_dirs[ThrusterMicroseconds.HEAVE_FRONT_LEFT])
-        pwm_arr[ThrusterMicroseconds.FRONT_LEFT] = force_to_pwm_thruster5(forces_msg.FRONT_LEFT * thruster_mount_dirs[ThrusterMicroseconds.FRONT_LEFT])
-        pwm_arr[ThrusterMicroseconds.FRONT_RIGHT] = force_to_pwm_thruster4(forces_msg.FRONT_RIGHT * thruster_mount_dirs[ThrusterMicroseconds.FRONT_RIGHT])
-        pwm_arr[ThrusterMicroseconds.HEAVE_FRONT_RIGHT] = force_to_pwm_thruster3(forces_msg.HEAVE_FRONT_RIGHT * thruster_mount_dirs[ThrusterMicroseconds.HEAVE_FRONT_RIGHT])
-        pwm_arr[ThrusterMicroseconds.HEAVE_BACK_RIGHT] = force_to_pwm_thruster2(forces_msg.HEAVE_BACK_RIGHT * thruster_mount_dirs[ThrusterMicroseconds.HEAVE_BACK_RIGHT])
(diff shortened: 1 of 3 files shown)


catkin_ws/src/propulsion/src/thrust_mapper_utils.py
```diff
@@ -19,61 +16,174 @@
...
+    - For forces above +0.5 N, a polynomial fit for positive forces is used.
+    - For forces below -0.5 N, a polynomial fit for negative forces is used.
+    - For small forces within ±0.5 N, the function linearly interpolates 
+      between 1500 µs and the polynomial curves to ensure smooth transitions 
+      and prevent jitter near zero thrust.
+    
+    The output PWM is clipped to remain within [1100, 1900] µs, which matches 
+    our ESC input limits.'''
+    
+    deadband_eps = 0.5
```
Alb33rt: This replaces the constant `DEADBAND_EPSILON` defined globally, however this should be an explicit constant as is with the old version. This should be applied to every function (which shouldn't be eight functions but generalized into one).

catkin_ws/src/propulsion/src/thrust_mapper.py
```diff
@@ -6,7 +6,17 @@
 
 import numpy as np
 import rospy
-from thrust_mapper_utils import *  # Assumes force_to_pwm, thruster_mount_dirs, etc. are defined here.
+from thrust_mapper_utils import (
```
Alb33rt: If importing everything from a file, use the asterisk to import everything (so no changed here).

catkin_ws/src/propulsion/src/thrust_mapper.py
```diff
@@ -115,14 +125,14 @@ def forces_to_pwm_publisher(self, forces_msg):
...
+        pwm_arr[ThrusterMicroseconds.HEAVE_BACK_LEFT] = force_to_pwm_thruster7(forces_msg.HEAVE_BACK_LEFT * thruster_mount_dirs[ThrusterMicroseconds.HEAVE_BACK_LEFT])
+        pwm_arr[ThrusterMicroseconds.HEAVE_FRONT_LEFT] = force_to_pwm_thruster6(forces_msg.HEAVE_FRONT_LEFT * thruster_mount_dirs[ThrusterMicroseconds.HEAVE_FRONT_LEFT])
+        pwm_arr[ThrusterMicroseconds.FRONT_LEFT] = force_to_pwm_thruster5(forces_msg.FRONT_LEFT * thruster_mount_dirs[ThrusterMicroseconds.FRONT_LEFT])
+        pwm_arr[ThrusterMicroseconds.FRONT_RIGHT] = force_to_pwm_thruster4(forces_msg.FRONT_RIGHT * thruster_mount_dirs[ThrusterMicroseconds.FRONT_RIGHT])
+        pwm_arr[ThrusterMicroseconds.HEAVE_FRONT_RIGHT] = force_to_pwm_thruster3(forces_msg.HEAVE_FRONT_RIGHT * thruster_mount_dirs[ThrusterMicroseconds.HEAVE_FRONT_RIGHT])
+        pwm_arr[ThrusterMicroseconds.HEAVE_BACK_RIGHT] = force_to_pwm_thruster2(forces_msg.HEAVE_BACK_RIGHT * thruster_mount_dirs[ThrusterMicroseconds.HEAVE_BACK_RIGHT])
+        pwm_arr[ThrusterMicroseconds.BACK_RIGHT] = force_to_pwm_thruster1(forces_msg.BACK_RIGHT * thruster_mount_dirs[ThrusterMicroseconds.BACK_RIGHT])
```
Alb33rt: There are eight different functions that have the same logic just with replaced coefficients for the polynomial. This should be generalized into only one single `force_to_pwm` function where the coefficients can be replaced in the parameter of the function. Something like `force_to_pwm(forces_msg.BACK_LEFT * thruster_mount_dirs[ThrusterMicroseconds.BACK_LEFT], coeffs_left: List, coeffs_right: List)`

Alb33rt (changes requested):
Overall this PR has correct functionality, so only stylistic changes that were mentioned require changes. Mainly, the `force_to_pwm` function should not be eight separate functions but one single function where the coefficients of the polynomial and be switched out via Arrays (Lists) in the parameter of the function. After these changes are done, it should be fine!

majdkhalife (approved):
lgtm

PR #692, opened 2025-05-08
Fixed function call in drytest

nadimasmar:
Fixed function call in drytest


JPGC04:
Approved but to be tested in terminal to make sure dry testing works

PR #693, opened 2025-05-09
Add drift test script for comparing old vs. new force-to-PWM mappings

nadimasmar:
This script performs an in-pool drift test to compare the performance of the old global force-to-PWM mapping against the new per-thruster polynomial mapping for our T200 thrusters.

Test procedure:

The AUV hovers at a constant depth using all vertical thrusters.

Forward motion is initiated for 5 seconds using front and rear thrusters.

The resulting lateral drift is measured physically (e.g., with a measuring tape or visual marker).

The same test is run for both the old and new PWM mappings, toggled via the use_old_coeffs flag in the script.

We compare the lateral drift between the two runs and confirm that our new thruster mappings are more accurate than the old ones

Key features:

Uses the existing ROS topic /propulsion/microseconds

Clean separation between hover and forward motion phases

Includes comments on the different indexing used by software and mechanical teams. This inconsistency leads us to have different thruster indices in the ThrusterMicroseconds message and the force_to_pwm functions. I handled the different indexing, but we should unify the indexing for simplicity in the future.

This script will help us validate whether the new mapping reduces unintended lateral drift caused by asymmetric thruster behavior.

### DIFF
--- a/catkin_ws/src/propulsion/src/Drift_Test.py
+++ b/catkin_ws/src/propulsion/src/Drift_Test.py
@@ -1,0 +1,113 @@
+#!/usr/bin/env python3
+
+"""
+Script to compare old and new force-to-PWM mappings.
+
+Procedure:
+1. Use the old mapping to dive and maintain constant depth.
+2. Move forward for 5 seconds using front and rear thrusters.
+3. Measure drift (e.g., with tape or visual markers).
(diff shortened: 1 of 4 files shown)


JPGC04:
Merged into a branch with same name to create separation from dockerIssues branch.

PR #696, opened 2025-05-13
Drift test script

nadimasmar:
Added drift test script


Alb33rt (approved):
Looks great and solid script!

PR #713, opened 2025-06-10
State estimationv1 WIP

nadimasmar:
**Summary**

This merge removes redundant code and introduces a streamlined, working control system for the AUV. The AUV can now perform basic movements and rotations using two core control methods, both of which have been extensively tested in simulation:

MoveDeltaLocal: Moves the AUV by specified displacements in the odometry (odom) frame.

Rotate: Rotates the AUV by specified angles about each of the three axes, following a Roll–Pitch–Yaw order.

In addition, a basic Line-of-Sight (LoS) tracker has been implemented. This module enables the AUV to follow a 3D spline trajectory with a defined lookahead distance, which allows executing simple missions like Pre-Qualification. However, this tracker has not yet been tested, due to issues in the sim that I showed on our Discord. 

Albert also added a lot of features and changes, he can describe in much better detail than I can. 

**Notes / Outstanding Issues**

State Estimation Dependency:
The control architecture assumes that the state estimation pipeline is functional and reasonably accurate. Until this is verified, real-world testing (e.g., in the pool) is not feasible.

SciPy Compatibility for LoS Tracker:
The LoS tracker relies on newer SciPy functions unavailable in the current AUV environment. We have two options:

Upgrade the SciPy version on the AUV

Modify the tracker to use compatible legacy functions

Simulation Bugs:
Due to unresolved issues in the sim, the LoS tracker has not been tested. Assistance with debugging the sim would be appreciated.


JPGC04:
There is something weird going on. The Docker Issues pr was squashed so github is not really recognizing some of these commits as duplicates. So the log is messed up in some way which we need to figure out how to resolve.

Alb33rt:
Someone needs to resolve these merge conflicts before we can review. I suspect these merge conflicts come from not squashing properly + unfamiliarity with git. That's totally alright but we need to clean it up or we are wasting time reviewing this.

PR #716, opened 2025-06-16
[WIP] - Controls and State Estimation

nadimasmar:
Branch Description:
I tried creating a cleaned-up version of the original stateEstimationv1 branch, squashed and rebased onto the latest noetic.
Unrelated history from other branches (e.g. dockerIssues) has been removed. Hopefully this cleans things up enough to start reviewing the changes we made. 

 Original branch preserved for history under stateEstimationv1. 
 
 
 **Summary**

This merge removes redundant code and introduces a streamlined, working control system for the AUV. The AUV can now perform basic movements and rotations using two core control methods, both of which have been extensively tested in simulation:

MoveDeltaLocal: Moves the AUV by specified displacements in the odometry (odom) frame.

Rotate: Rotates the AUV by specified angles about each of the three axes, following a Roll–Pitch–Yaw order.

In addition, a basic Line-of-Sight (LoS) tracker has been implemented. This module enables the AUV to follow a 3D spline trajectory with a defined lookahead distance, which allows executing simple missions like Pre-Qualification. However, this tracker has not yet been tested, due to issues in the sim that I showed on our Discord.

Albert also added a lot of features and changes, he can describe in much better detail than I can.

**Notes / Outstanding Issues**

State Estimation Dependency:
The control architecture assumes that the state estimation pipeline is functional and reasonably accurate. Until this is verified, real-world testing (e.g., in the pool) is not feasible.

SciPy Compatibility for LoS Tracker:
The LoS tracker relies on newer SciPy functions unavailable in the current AUV environment. We have two options:

Upgrade the SciPy version on the AUV

Modify the tracker to use compatible legacy functions

Simulation Bugs:
Due to unresolved issues in the sim, the LoS tracker has not been tested. Assistance with debugging the sim would be appreciated.


catkin_ws/src/planner/src/Prequal.py
```diff
@@ -0,0 +1,71 @@
+#!/usr/bin/env python3
+import rospy
+from substates.utility.controller import Controller
+import numpy as np
+from scipy.interpolate import make_splprep
```
nadimasmar: This needs a SciPy version ≥ 1.11.0. We should either update our Scipy version or we can change the code to use legacy functions.

catkin_ws/src/propulsion/src/thrust_mapper.py
```diff
@@ -1,98 +1,96 @@
...
-        return Wrench(force=Vector3(*force_body), torque=Vector3(*torque_body))
-
-
     def wrench_to_thrust(self, wrench_msg):
         """
         Callback function that maps a received Wrench message into thruster forces.
         It first converts the wrench from global to body frame, then applies the
-        pseudo-inverse of the thruster mapping matrix.
+        pseudo-inverse of the thruster mapping matrix. We assume that all messages published on /controls/effort
+        are in the global frame.
```
nadimasmar: Small error, "global" should be replaced with "base_link"

catkin_ws/src/planner/src/Prequal.py
majdkhalife: Change the code to use legacy functions

catkin_ws/src/propulsion/src/thrust_mapper.py
majdkhalife: its a comment, doesnt matter lol.

majdkhalife (dismissed):
just make change to use legacy function and if you want fix the typo ow lgtm

nadimasmar:
Changed LOS tracker to use legacy functions. Also tested the tracker in the sim, it works.

catkin_ws/src/planner/src/substates/utility/controller.py
JPGC04, line 1: Are we fully abandoning this file?

catkin_ws/src/planner/src/substates/utility/yaw_controller.py
JPGC04, line 1: Does the existence of a yaw controller allude to having a roll and pitch controller in the future? Or do we not care about the DOF for our missions?

catkin_ws/src/sensors/src/dvl_republish.cpp
```diff
@@ -32,10 +32,6 @@ int main(int argc, char **argv)
     // Advertise the _with_covariance message
     pub_dvl = nh.advertise<geometry_msgs::TwistWithCovarianceStamped>("/sensors/dvl/twist", 10);
 
-    // Subscribe to your raw DVL velocities
```
JPGC04, line 35: By doing the deletion 36 and 37, you render the callback function rawDvlCb on line 8 useless.
In other words, this file effectively does nothing as a whole.

catkin_ws/src/sensors/src/dvl_convariance.py
JPGC04, line 1: All of this means the DVL data will not be stamped no? Unless this is handled in another file?

JPGC04 (dismissed):
I still need to do a sanity check in the sim to make sure, pwm is still being published, but otherwise I am okay with approving, and will approve to avoid a block.

JPGC04:
I did a sanity check and dry tests still work, and when running pooltest.launch, the pwm values are sensical for what the python file is trying to achieve. Irl we will need clean sensor data and pid tuning, but I am very happy with this PR and it demonstrates a functional PID.
The only blocker we have is that there is some depedency issues in the sim, specifically scipy is not installed at all in some Docker images of the sim (I was playing around with this). I will continue working on this on its respective branch. A similar issue exists for torch and potentially for scikit-learn, but these issues are to be resolved in their respective branches.
@nadimasmar @majdkhalife

catkin_ws/src/planner/src/substates/utility/controller.py
nadimasmar: Not at all. This file contains all our functions control methods. It was not rendered because there is a large diff, not because it was deleted

catkin_ws/src/planner/src/substates/utility/yaw_controller.py
nadimasmar: Yaw controller is a redundant file. We control Roll, Pitch, and Yaw with our rotate() methods and can maintain constant heading with the quaternion PID. I was willing to delete it but I dont know if other people have use for it. I'm not sure it even works. I would like to delete it with @majdkhalife or @Alb33rt approval.

nadimasmar: In addition, the effort and quaternion servers are redundant as well. The new control methods in controller.py directly publish setpoints for X,Y,Z and rotation on their respective topics. The servers are not used. We can keep the servers in case we use them in the future. They can also be removed for simplicity. I wouldn't delete the servers without double-checking they have no future use.

catkin_ws/src/sensors/src/dvl_convariance.py
nadimasmar: This deletion was part of our efforts on working on sensors and state estimation. Albert was the one who deleted this file, I don't recall why he deemed it to be redundant. I say we keep this change as the controls model is working in the way things are. However, all future work on sensors and/or state estimation should be done in a new branch.

catkin_ws/src/sensors/src/dvl_republish.cpp
nadimasmar: Yes, this file isnt doing anything. Perhaps it should be deleted as well. For consistency, I say we keep this change to avoid breaking anything elsewhere, and look into it and its potential removal in a branch dedicated to sensors and state estimation.

JPGC04: If you want to keep the file but you want to stop it from doing anything, please comment stuff out instead of deleting.
I have two questions:
1. Are we gonna use twist or pose?
2. Do we want the convariances to be stamped on the data?

catkin_ws/src/sensors/src/dvl_convariance.py
JPGC04: I am concerned about wanting to remove convariance. I am wondering whether I am misunderstanding this cause Albert explained to me that the convariances were hardcoded (bad) which might explain why he wants to delete this file.

catkin_ws/src/sensors/src/dvl_republish.cpp
nadimasmar: 1. We are going to use twist from the DVL.
2. Both our IMU and DVL give covariances with every reading. So yes, we want to stamp the data with covariance.

catkin_ws/src/sensors/src/dvl_convariance.py
nadimasmar: A new dvl_covariance file has been written that stamps incoming DVL data with dynamic covariances.

catkin_ws/src/sensors/src/dvl_republish.cpp
nadimasmar: both dvl_republish and imu_reublish have been deleted. dvl_covariance.py and imu_covariance.py now handle republishing and covariance stamping.

nadimasmar:
What's left is to debug the reworked state estimation logic in the sim before moving on to pool tests.

PR #702, opened 2025-05-25
Docker cleaning and workflow push to dockerhub

JPGC04:
Cleaning up Dockerfile and making sure workflows only dispatch on changes (will still dispatch on schedule).

PR #717, opened 2025-06-18
Updated Dockerfile to not update 4 times. Should lead to faster builds. 

majdkhalife:


JPGC04 (changes requested):
I want this PR closed not merged since in the dockerWorkflow PR #702 it takes care of sim and cleans it fully.

PR #726, opened 2025-07-26
Changes to the Vision package

llandauro:
## Changes Made
- image_collection.launch: modified to directly subscribe to the Zed camera's messages (no bridge anymore)
- image_collection.py: updated the file with the correct subscribers, set up image collection user flow and worked on UX for ease of use
- Dockerfile (ros-1-douglas): incorporated fixes to the scikit learn problem (cannot allocate memory in static TLS block) and the installation of the dill module, manually installed a torchvision pkg and set up scikit crashes fix automatically
- vision.py: made is_vision_ready True to debug and printed names of detections in vision.py
- vision.launch: no changes so far, more to come

## Rationale
The vision package is sturdy enough and has been tested multiple times in the pool.
 
## Tests Done
- Wifi and no Wifi front camera and down camera testing
- Pool testing without the thrusters activated


JPGC04 (commented):
You might need to branch into state estimation squashed branch. im not sure where this branch branched of from initially.
Also, you need to update this branch with noetic (aka merge noetic into your branch)

catkin_ws/src/ZED_Diagnostic_Results.json
Rain1618, line 1: Do we need this file??

llandauro: I don't think so??

PR #728, opened 2025-07-31
Zed ros wrapper

JPGC04:


thirteenlizards (approved):
slay

asiancooki (approved):

.github/workflows/catkin_build.yml
```diff
@@ -0,0 +1,44 @@
...
+        uses: actions/checkout@v3
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Build base image
+        uses: docker/build-push-action@v5
+        with:
+          context: Docker/base
+          file: Docker/base/Dockerfile
```
Alb33rt, line 28: is there a reason the `/base` is the one being tested for on ARM instead of the `/ros1` Dockerfile?

.github/workflows/catkin_build.yml
```diff
@@ -0,0 +1,44 @@
...
+
+      - name: Run catkin build inside container
+        run: |
+          docker run --rm \
+            -v ${{ github.workspace }}:/ws \
+            -w /ws/catkin_ws \
+            local-base-image:latest \
+            bash -c "source /opt/ros/noetic/setup.bash && \
+                    cd src && \
+                    rm -rf zed-ros-* && \
```
Alb33rt, line 42: Are we removing `/zed_ros_wrapper` during the build?

Alb33rt (commented):
I have two questions on the way the CI/CD is setup. Other than that everything else lgtm. I'll approve soon after I clear up my confusion.

.github/workflows/catkin_build.yml
JPGC04: We have historically always used base for ci/cd in github actions. using ros1 could work, but historically this has never been the case, i am totally willing to change this.

.github/workflows/catkin_build.yml
JPGC04: In consequence to the fact that we use base in github actions, zed ros wrapper cannot build (lack of cuda and lack of zed sdk)

.github/workflows/catkin_build.yml
Alb33rt: Most of the changes we do in `/ros1` are going to be important in `/base` as well. The `/base` Dockerfile is always going to look different from the one used on the jetson. I think we can look into deploying two different CIs.

.github/workflows/catkin_build.yml
Alb33rt: Makes sense considering we're build on base. I had the assumption that we were doing this for `/ros1`.

PR #733, opened 2025-08-04
ZED Integration in Base + Increase Scheduled Workflows

JPGC04:


Alb33rt (approved):
LGTM. Using manifests is interesting. I don't think the efficiency is that important prior to competition, so it's probably okay.

asiancooki (approved):

PR #738, opened 2025-08-11
Clean frontcam

llandauro:
## Related Issue(s)
  - The Zed camera was not working
  - vision.launch was not launching properly due to TLS stack memory issues<
  - vision.launch crashed on detections due to torchvision not being installed properly
  - The Image Collection was tedious to do

## Changes Made
- Zed camera implementation and vision. launch works
- Improved image collection user flow
- Added ENV variable LD_PRELOAD 
- Reordered the import statements in object_detection (required for some reason)
- Imported packages dill to load YOLO model
- Made torchvision install on the first docker build, otherwise the install was not recognized

- Research shows the TLS stack memory is fixed with LD_PRELOAD and reordering of imports
- Missing Dill module to implement the Zed camera
- The current user flow failed easily, which made us lose time

## Tests Done
- Tested in MEDN with and without wifi
- Tested at the pool, without wifi

### DIFF
--- a/Docker/ros1/Dockerfile
+++ b/Docker/ros1/Dockerfile
@@ -89,28 +89,39 @@ RUN pip install --no-cache-dir -r requirements.txt \
     && pip install --no-deps --no-cache-dir ultralytics==8.0.207 \
     && rm -f /usr/local/bin/cmake && rm -rf /root/.cache/pip && rm requirements.txt
 
-RUN git clone --branch release/0.16 --recursive --depth=1 https://github.com/pytorch/vision torchvision && \
-    cd torchvision && \
-    git checkout release/0.16 && \
-    python3 setup.py bdist_wheel && \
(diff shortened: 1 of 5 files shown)


catkin_ws/src/vision/launch/vision.launch
majdkhalife, line 1: maybe remove the comments 🧐. Also they would beed to match the classes on rf

majdkhalife (approved):
Im good witg everything, @JPGC04 the docker fix is kinda cheap but it works so lmk if you wanna include it. I think we should for comp and then after get a more perma fix

Docker/ros1/Dockerfile
JPGC04, line 1: If it passes checks I am okay with it. It doesn't seem like the cleanest implementation, but this has been tested so its good. To be cleaned after comp.

JPGC04 (approved):

PR #739, opened 2025-08-13
jetson docker fix

JPGC04:


JPGC04:
Already merged into noetic

PR #740, opened 2025-08-15
Merge State Estimation and Republishers Rework into noetic

dougieauv:
## Changes Made
- Configured `robot_localization`' setup to use double Extended Kalman Filter (EKF)s, allowing for publishing of separate `map -> odom` and `odom -> auv` transformations.
- Added the pose from the ZED2i camera into our EKF setup. Outputs good /state/pose even without DVL data.
- Reworked sensor republishers from Python to C++, significantly improving lowering the delay of data prior to inputting to the EKF.
- Decreased the drift in the estimated pose by adjusting frequencies and physical system parameters in the EKF nodes.
- `zed_ros_wrapper` added to enable out-of-the-box usage of GraphSLAM from the ZED SDK.
- `xsens_mti_ros_driver` added and configured for high frequency IMU data publishing into the ROS stack.

- State estimation clarifications were made after researching more into GraphSLAM. The old `robot_localization` setup does not reflect [REPL-103](https://www.ros.org/reps/rep-0103.html) compliant navigation schematics.
- State estimation drifts significantly without the use of a DVL. A global source of truth was required underwater. Therefore, a GraphSLAM library was required.

## Tests Done
- Attached ZED2i camera and Movella MTi-630 IMU to a bowl and moved them around as a unit. Orientation and absolute position reflects real-world positioning when outside of water.

## Optional Screen Recording
- Made at competition (recordings later).

- `State Estimation pages need to be reconfigured to reflect the setup of the dual EKF.

- Changed reflected in Dockerfile.

- `stateEstimationv1-squashed`
- `front-cam`


catkin_ws/src/bringup/launch/bringup.launch
Rain1618, line 1: why do we need the display here?

catkin_ws/src/controls/launch/pid.launch
Rain1618, line 1: pid values should not be changed unless tuned in pool, we should not be merging changes related to controls yet since they haven't been tested in real life

catkin_ws/src/controls/src/quaternion_pid.py
Rain1618, line 1: we should see if we can get the rate higher, else im chill with merging this file after testing

catkin_ws/src/orb_slam3_ros
Rain1618, line 1: we're not using orb slam rn, no need to merge this yet

catkin_ws/src/propulsion/src/thrust_mapper_utils.py
Rain1618, line 1: why are we doing 1*force?

catkin_ws/src/sensors/src/dvl_covariance.cpp
Rain1618, line 1: i dont think this file has ever been tested...

catkin_ws/src/sensors/src/zed_pose_republish.py
Rain1618, line 1: if we are using the cpp file, delete this one

catkin_ws/src/bringup/launch/bringup.launch
dougieauv: We changed unity bridge to now publish float64 values as depth. As a result, we need the depth republisher to republish this data as a PosewithCovarianceStamped message. This makes the sim a bit more closer to our real life sensors --> state estimation pipeline. 
- Nadim

JPGC04 (dismissed):

catkin_ws/src/controls/launch/pid.launch
dougieauv: sure, but theyre outdated anyways might as welkl have neotic be more stable and up to date withg what we do

catkin_ws/src/controls/src/quaternion_pid.py
dougieauv: noetic old, talk to me irl
- Majd

catkin_ws/src/sensors/src/zed_pose_republish.py
dougieauv: Sure,

majdkhalife (dismissed):
noetic needs to be updated. Its stable and matches what we did in comp.  iff we had the old PID values. Oh well

majdkhalife (approved):
this should be approved as it was what we used at comp. Also we should never have branches this ahead of noetic ever. no more dockerIssues nor stateEstimationv1

Rain1618 (approved):

